PR #29, opened 2025-12-10
Fix stream authentication protocol implementation

KyleAMathews:
The protocol previously suggested using Cache-Control: private and
Vary: Authorization for authenticated streams. This was problematic:

1. Most CDNs ignore Vary header (except Accept-Encoding)
2. Cache-Control: private disables CDN caching entirely

The correct approach (following Electric SQL's auth patterns):
- Use an authorizing proxy that validates credentials BEFORE the CDN
- Cache keys are based on resource (URL, offset), not user identity
- Multiple authorized users share cached responses for same resource
- Proxy adds Vary: Authorization for BROWSER cache isolation between login sessions (not CDN isolation)

Changes:
- Section 8.1: Clarify Vary is for browser cache isolation only, CDNs ignore it. Proxy should add it for session isolation.
- Section 8.3: Add comprehensive auth/caching guidance including:
  - Proxy authentication pattern
  - Gatekeeper authentication pattern (shape-scoped tokens)
  - Cache architecture diagram
  - Note about proxy adding Vary header
  - Guidance for user-specific data
- Section 10.1: Update cross-reference to recommend proxy pattern

References:
- Electric SQL auth guide: https://electric-sql.com/docs/guides/auth

Fixes: #27

### DIFF
--- a/PROTOCOL.md
+++ b/PROTOCOL.md
@@ -594,13 +594,13 @@ When streams require authentication and CDN caching, implementers should conside
 
 #### Embedded Authentication
 
-When streaming logic is embedded in your application server, authentication can be handled directly using your application's existing auth mechanisms. The application validates credentials and serves stream data in a single request path:
+When streaming logic is embedded in your application server, authentication can be handled directly using your application's existing auth mechanisms:
 
 ```
-Client → CDN → Application Server (auth + streaming)
+Client → Application Server (auth + streaming)
 ```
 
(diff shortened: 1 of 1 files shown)


KyleAMathews:
Hmmm the cache headers people should respond with is definitely in scope for the protocol — but yes specific advice then about auth isn't — I'll move it to an appendix.

KyleAMathews:
@samwillis ready for review again

Issue #32, opened 2025-12-11
Optimistic Concurrency Control (OCC) Support


yordis:
Hey team, I love to see this type of projects going on. I wonder, for the `append` operation, are you intending to add optimistic concurrency control (`If-Match` HTTP header)?

Issue #1, opened 2025-12-05
Add support for 'offset=now' to skip historical data


KyleAMathews:
## Problem

The protocol currently has no way to start following a stream from "now" without first fetching all historical data. This is critical for use cases where you only care about new events, not the full history.

### Current Limitations

1. GET without `offset` → reads from stream start (offset 0)
2. Long-poll requires an offset, forcing clients to catch up first
3. No way to "jump to tail" and start receiving only new data

### Critical Use Cases

**Presence APIs**
```javascript
// I want to know who's online NOW, not replay every join/leave event
stream.follow({ fromNow: true })
```

**Live Monitoring/Alerts**
```javascript
// Start receiving new alerts immediately, don't replay old ones
const alerts = new DurableStream({ url: '/alerts' })
for await (const alert of alerts.follow({ fromNow: true })) {
  notify(alert)
}
```

**Late Joiners**
```javascript
// Join an active conversation without loading full history
chat.follow({ fromNow: true })
```

## Solution: Follow Electric's Pattern

Electric SQL implements `offset=now` as a special sentinel value, and it works well. We should adopt the same pattern.

### Electric's Implementation

```elixir
# packages/sync-service/lib/electric/replication/log_offset.ex
def from_string("now"), do: {:ok, :now}

# packages/sync-service/lib/electric/shapes/api.ex
defp seek(%Request{params: %{offset: :now}} = request) do
  # Returns immediately with up-to-date message and last_offset
  request
  |> use_last_offset_as_chunk_end()
  |> set_response_offset_for_now()
end
```

Returns immediately with:
- `up-to-date: true`
- `Stream-Next-Offset` set to current tail
- Empty body (no historical data)

### Proposed API

```http
GET /stream?offset=now
```

**Behavior:**
- Returns 200 with empty body
- `Stream-Next-Offset` header = current tail offset
- `Stream-Up-To-Date: true`
- Client can then start long-polling from that offset

**Protocol spec addition:**
```markdown
### Special Offset Values

Offsets are opaque tokens, with the following special values:

- **Omitted**: Read from stream start (offset 0)
- **`now`**: Skip to current tail without transferring historical data
  - Returns 200 with empty body
  - Sets `Stream-Next-Offset` to current tail
  - Sets `Stream-Up-To-Date: true`
  - Useful for presence, monitoring, and late-joiner scenarios
  - Can be combined with `live=long-poll` to start live tailing immediately
```

### Why This Works

[...]

PR #34, opened 2025-12-11
Add support for optimistic concurrency control (OCC) with the If-Match header on POSTs

KyleAMathews:
## Summary

Adds support for the standard HTTP `If-Match` header on POST (append) requests to enable optimistic concurrency control (OCC). This allows multiple clients to safely write to the same stream without lost updates—one succeeds, others get 412 and can retry.

---

### Approach

Uses standard HTTP conditional request semantics (RFC 9110) rather than inventing a custom header. The flow:

1. Client reads stream, gets `Stream-Next-Offset: abc123`
2. Client appends with `If-Match: abc123`
3. If stream hasn't changed → 204 success
4. If another writer appended → 412 with current offset for retry

Key implementation detail: The `FileBackedStreamStore` uses per-stream locks (`streamLocks` Map) to ensure the offset check + write is atomic:

```typescript
// Acquire per-stream lock to serialize appends
const existingLock = this.streamLocks.get(streamPath) ?? Promise.resolve()
// ... check expectedOffset, then write ...
```

The in-memory `StreamStore` is already single-threaded so atomicity is implicit.

### Key Invariants

1. **If-Match evaluated before Stream-Seq** — Protocol specifies evaluation order; 412 takes precedence over 409
2. **412 returns current offset** — Enables client retry without extra HEAD request
3. **Quoted and unquoted values accepted** — RFC 9110 says ETags should be quoted, but we accept both for compatibility
4. **Wildcard `*` means "stream must exist"** — Returns 412 on non-existent stream (not 404)

### Non-goals

- **Write-side ETag not implemented** — Protocol says servers MAY return `ETag` header on append success; we rely on `Stream-Next-Offset` which serves the same purpose
- **No client-side OCC helpers** — This is server-only; client can use raw fetch with If-Match

### Trade-offs

**Alternative considered: Custom `Stream-Expected-Offset` header**
- Pros: Clearer purpose, no RFC baggage
- Cons: Reinventing the wheel, less interoperable
- Decision: Use standard `If-Match` for HTTP semantics alignment

**Alternative considered: Lock-free CAS in file store**
- Pros: Higher throughput under contention
- Cons: Complex, LMDB already serializes writes anyway
- Decision: Simple per-stream locks; contention is rare in practice

---

## Verification

```bash
# Run all tests (includes 18 new OCC conformance tests)
pnpm test:run

# Key test scenarios covered:
# - Correct If-Match succeeds
# - Mismatched offset returns 412
# - Concurrent writers: one wins, one gets 412
# - Quoted/unquoted values both work
# - If-Match: * on non-existent stream returns 412
# - If-Match evaluated before Stream-Seq
```

---

## Files Changed

| File | Change |
|------|--------|
| `PROTOCOL.md` | Added If-Match header spec, 412 response code, OCC security section (10.7) |
| `packages/server/src/store.ts` | Added `OffsetMismatchError`, `expectedOffset` option to append() |
| `packages/server/src/file-store.ts` | Added `streamLocks` for atomic OCC, `expectedOffset` handling |
| `packages/server/src/server.ts` | Parse If-Match header, handle 412 responses, `parseIfMatchValue()` |
| `packages/server-conformance-tests/src/index.ts` | 18 new OCC test cases |

---

Addresses: https://github.com/durable-streams/durable-streams/issues/32

### DIFF
--- a/PROTOCOL.md
+++ b/PROTOCOL.md
@@ -175,7 +175,7 @@ Servers that do not support appends for a given stream **SHOULD** return `405 Me
   - `Stream-Seq` values are opaque strings that **MUST** compare using simple byte-wise lexicographic ordering. Sequence numbers are scoped per authenticated writer identity (or per stream, depending on implementation). Servers **MUST** document the scope they enforce.
   - If provided and less than or equal to the last appended sequence (as determined by lexicographic comparison), the server **MUST** return `409 Conflict`. Sequence numbers **MUST** be strictly increasing.
 
-- `If-Match: <offset>` (optional)
+- `If-Match: <offset>`
   - Enables optimistic concurrency control (OCC) for the append operation.
   - The value **MUST** be the expected current tail offset of the stream (i.e., the `Stream-Next-Offset` value from a previous response).
   - The value **MAY** be quoted (e.g., `If-Match: "abc123"`) or unquoted (e.g., `If-Match: abc123`). Servers **MUST** accept both formats by stripping surrounding quotes when comparing. Quoted form is recommended per RFC 9110.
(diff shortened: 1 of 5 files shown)


yordis (approved):

KyleAMathews:
At its core, durable streams is a just a pure byte stream so questions like "how can I tell how far I am in the log" don't entirely make sense. Offsets are just an implementation detail so we don't want them to have inherit meaning e.g. they could be a byte offset or anything else (perhaps related to the underlying storage). The protocol just says if you return the protocol you got on the last request, you're guaranteed to get the next series of bytes in the log.

> In this case, when I append and you return Stream-Next-Offset, I can pass it to clients so they can determine whether they got the exact same message or catch up faster through a websocket or another HTTP connection or deal with out of order messaging, optimistic UI updates, and so on.
You can also use Stream-Next-Offset to query the database for a projection that’s at least up to that offset.

Yeah you can do stuff like this but I think also important to note is we don't support range queries — you just start consuming the log from the start and keep going. We will support starting from the current point of the log as that's useful sometimes but nothing else is meaningful within this system unless you store offsets and make meaning elsewhere https://github.com/durable-streams/durable-streams/issues/1

yordis:
> important to note is we don't support range queries — you just start consuming the log from the start and keep going.

Will your spec be friendly to actual Event Sourcing pattern needs? That is why I asked.

KyleAMathews:
I'm not really sure what you're asking... the offset ensures you don't miss messages. I'm not sure why a producer would need or want to communicate offsets directly to clients — the clients would get the update probably before the producer could send it. Perhaps start an issue where we discuss things in more depth?

KyleAMathews:
The protocol guarantees you'll get every message. If you have 1 and ask for more, the server will get you 2..n.

PR #37, opened 2025-12-11
Add conformance tests for SSE

KyleAMathews:
- Add 9 SSE conformance tests covering:
  - SSE content-type response (text/event-stream)
  - SSE for application/json streams
  - Offset parameter requirement for SSE mode
  - Client rejection of SSE for incompatible content types
  - Data event streaming
  - Control events with offset
  - Cursor parameter support
  - JSON array wrapping
  - Empty stream handling

- Implement SSE mode in server:
  - Handle live=sse query parameter
  - Return text/event-stream content-type
  - Send data events for each message
  - Send control events with offset/cursor
  - Support keep-alive on timeout

- Fix server shutdown with active SSE/long-poll connections:
  - Add cancelAllWaits() to StreamStore and FileBackedStreamStore
  - Track active SSE responses and close them on shutdown
  - Check isShuttingDown flag in SSE loop

### DIFF
--- a/packages/conformance-tests/src/index.ts
+++ b/packages/conformance-tests/src/index.ts
@@ -19,6 +19,61 @@ export interface ConformanceTestOptions {
 }
 
 /**
+ * Helper to fetch SSE stream and read until a condition is met.
+ * Handles AbortController, timeout, and cleanup automatically.
+ */
+async function fetchSSE(
+  url: string,
+  opts: {
+    timeoutMs?: number
+    maxChunks?: number
+    untilContent?: string
+    headers?: Record<string, string>
+  } = {}
+): Promise<{ response: Response; received: string }> {
+  const { timeoutMs = 2000, maxChunks = 10, untilContent, headers = {} } = opts
+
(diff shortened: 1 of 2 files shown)


packages/conformance-tests/src/index.ts
```diff
@@ -2186,6 +2186,698 @@ export function runConformanceTests(options: ConformanceTestOptions): void {
...
+
+      try {
+        const response = await fetch(
+          `${getBaseUrl()}${streamPath}?offset=-1&live=sse`,
+          {
+            method: `GET`,
+            headers: { Accept: `text/event-stream` },
+            signal: controller.signal,
+          }
+        )
```
samwillis: This pattern with a fetch with a timeout and then reaving a little bit of the body for a while is used repeatedly through the tests. It may be cleaner to make a little helper that abstracts it away.
Each use is a little different though so needs a little consideration.

samwillis (approved):
Approving as the tests are good, but with two thoughts that you may want to address first.

Issue #42, opened 2025-12-12
Spec deviation: section 5.7 - SSE control message


broady:
Spec says:

````
### 5.7. Read Stream - Live (SSE)

#### Response Format

Data is emitted in [Server-Sent Events format](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events#event_stream_format).

**Events:**

- `data`: Emitted for each batch of data
  - Each line prefixed with `data:`
  - When the stream content type is `application/json`, implementations **MAY** batch multiple logical messages into a single SSE `data` event by streaming a JSON array across multiple `data:` lines, as in the example below.
- `control`: Emitted after every data event
  - **MUST** include `streamNextOffset` and **MAY** include `streamCursor`
  - Format: JSON object with offset and optional cursor. Field names use camelCase: `streamNextOffset` and `streamCursor`.

**Example:**

```
event: data
data: [
data: {"k":"v"},
data: {"k":"w"},
data: ]

event: control
data: {"streamNextOffset":"123456_789","streamCursor":"abc"}
```

#### Connection Lifecycle

- Server **SHOULD** close connections roughly every ~60 seconds to enable CDN collapsing
- Client **MUST** reconnect using the last received `streamNextOffset` value from the control event
````

Conformance test:

```
        // Verify control event format
        expect(received).toContain(`event: control`)
        expect(received).toContain(`Stream-Next-Offset`)
```

Implementation:

```
// Protocol headers (aligned with PROTOCOL.md)
const STREAM_OFFSET_HEADER = `Stream-Next-Offset`
const STREAM_CURSOR_HEADER = `Stream-Cursor`
```

```
      // Compute offset the same way as HTTP GET: last message's offset, or stream's current offset
      const controlOffset =
        messages[messages.length - 1]?.offset ?? stream!.currentOffset

      // Send control event with current offset/cursor
      const controlData: Record<string, string> = {
        [STREAM_OFFSET_HEADER]: controlOffset,
      }
      if (cursor) {
        controlData[STREAM_CURSOR_HEADER] = cursor
      }

      res.write(`event: control\n`)
      res.write(encodeSSEData(JSON.stringify(controlData)))
```

PR #43, opened 2025-12-12
Update SSE implementation to match spec for control message

broady:
Spec says streamNextOffset/streamCursor but implementation was Stream-Next-Offset/Stream-Cursor

Updates #42.


broady:
I'll rebase. Gimme a bit.

KyleAMathews:
Thanks!

PR #44, opened 2025-12-12
Fix Content-Type conformance test

broady:
The existing code literally sent

	Content-Type: undefined

over the wire.


broady:
thanks, missed that

PR #45, opened 2025-12-12
@durable-streams/state package and state-db (using Tanstack DB)

samwillis:
fork of `@durable-streams/state` branch #30, rebased and stacked on #38

implements the Tanstack DB based state-db.


KyleAMathews:
Merged this into #30

PR #51, opened 2025-12-14
Exactly-once writes with Idempotent Producers

KyleAMathews:
For production use, wrap your stream with `IdempotentProducer` to get exactly-once delivery guarantees with automatic retries:

```typescript
import { DurableStream, IdempotentProducer } from "@durable-streams/client"

const stream = await DurableStream.create({
  url: "https://your-server.com/v1/stream/my-stream",
  contentType: "application/json",
})

// Create an idempotent producer with pipelining for high throughput
const producer = new IdempotentProducer({
  stream,
  maxBatchBytes: 1024 * 1024, // Flush when batch reaches 1MB (default)
  maxInFlight: 5, // Up to 5 batches in-flight concurrently (default)
  onBatchAck: (result, count) => {
    metrics.increment("events_sent", count)
  },
  onError: (error, count) => {
    logger.error("Batch failed", { error, count })
    // Could dead-letter, alert, etc.
  },
})

// Fire-and-forget - batched automatically, retried on failure
producer.append({ event: "click", button: "submit" })
producer.append({ event: "click", button: "cancel" })
producer.append({ event: "page_view", page: "/home" })

// Wait for all pending to complete (e.g., before shutdown)
await producer.flush()
console.log(`All sent, last acked: ${producer.state.lastAckedSequence}`)
```

**Why use IdempotentProducer?**

Network failures create uncertainty - when a request times out, you don't know if the server received it. Without idempotency, retrying could write the same data twice. The idempotent producer:

- **Batches** multiple appends and flushes when byte threshold is reached
- **Pipelines** up to 5 batches in-flight with Kafka-style client-side ordering
- **Retries** failed batches with exponential backoff (including out-of-order)
- **Tracks** sequence numbers per batch for duplicate detection
- **Ignores** duplicate sequences on the server (safe retries)
- **Fences** zombie producers via epochs (prevents stale producers from writing)
- **Provides** `onBatchAck` and `onError` callbacks for monitoring
- **Flushes** via `flush()` method to wait for all pending writes

Use `IdempotentProducer` when data integrity matters (payments, orders, events). Use simple `append()` for development, testing, or when duplicates are acceptable at the application level.

### DIFF
--- a/README.md
+++ b/README.md
@@ -194,11 +194,9 @@ const stream = await DurableStream.create({
   contentType: "application/json",
 })
 
-// Create an idempotent producer with pipelining for high throughput
+// Create an idempotent producer
 const producer = new IdempotentProducer({
   stream,
-  maxBatchBytes: 1024 * 1024, // Flush when batch reaches 1MB (default)
-  maxInFlight: 5, // Up to 5 batches in-flight concurrently (default)
   onBatchAck: (result, count) => {
     metrics.increment("events_sent", count)
   },
@@ -222,14 +220,15 @@ console.log(`All sent, last acked: ${producer.state.lastAckedSequence}`)
 
 Network failures create uncertainty - when a request times out, you don't know if the server received it. Without idempotency, retrying could write the same data twice. The idempotent producer:
 
-- **Batches** multiple appends and flushes when byte threshold is reached
-- **Pipelines** up to 5 batches in-flight with Kafka-style client-side ordering
-- **Retries** failed batches with exponential backoff (including out-of-order)
-- **Tracks** sequence numbers per batch for duplicate detection
-- **Ignores** duplicate sequences on the server (safe retries)
+- **Low latency** - Messages send immediately when no requests are in-flight
+- **Batches opportunistically** - Groups messages when requests are already in-flight
+- **Retries** failed requests with exponential backoff
+- **Tracks** sequence numbers for duplicate detection
+- **Ignores** duplicates on the server (safe retries)
 - **Fences** zombie producers via epochs (prevents stale producers from writing)
 - **Provides** `onBatchAck` and `onError` callbacks for monitoring
-- **Flushes** via `flush()` method to wait for all pending writes
+
+For high-throughput bulk loading, you can tune `maxBatchBytes` (flush threshold, default 1MB) and `maxInFlight` (concurrent batches, default 5). But for typical small/infrequent messages, the defaults prioritize low latency.
 
 Use `IdempotentProducer` when data integrity matters (payments, orders, events). Use simple `append()` for development, testing, or when duplicates are acceptable at the application level.
 
--- a/packages/client/src/idempotent-producer.ts
+++ b/packages/client/src/idempotent-producer.ts
@@ -7,7 +7,8 @@
  * - Sequence number tracking with duplicate detection
  * - Client-side ordering - batches are sent and acknowledged in strict sequence order
  * - Safe retries - duplicates are detected and ignored
- * - Automatic batching for high throughput
+ * - Low-latency default - messages send immediately when no requests are in-flight
+ * - Opportunistic batching - groups messages only when requests are already pending
  * - Automatic retry with exponential backoff
  */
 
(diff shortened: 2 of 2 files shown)


PROTOCOL.md
```diff
@@ -193,6 +194,199 @@ Servers that do not support appends for a given stream **SHOULD** return `405 Me
...
+
+```http
+POST {stream-url}
+Content-Type: application/json
+Stream-Producer-Id: ?
+
+[{"event": "created"}]
+```
+
+The special value `?` requests the server to assign a new producer ID. The server responds with:
```
kevin-dp: What if this `POST` request times out? Then we have a similar problem because we don't know if the server has received the request or not.

We cannot simply retry this `POST` request because if it succeeded we will have duplicate data because we don't know the producer ID and so now we appended twice and got different producer IDs.

I believe this only works if either the user generates a unique producer ID or the server assigns a unique producer ID but this requires an extra round trip at the beginning in order to get the producer ID before we can append.

So basically, you wouldn't be able to create a stream with initial data. You would create a stream, get a producer ID back, and then append the initial data. If the stream creation fails you can safely retry because stream creation is idempotent since it is based on a unique `stream-url` which the server uses to detect that the stream already exists.

So, stream creation already uses a client-generated unique ID in order to be idempotent. Would make sense to me that we also use a client-generated producer ID (+ seq number and epoch) to make appends idempotent.

PROTOCOL.md
```diff
@@ -193,6 +194,199 @@ Servers that do not support appends for a given stream **SHOULD** return `405 Me
...
+```http
+POST {stream-url}
+Content-Type: application/json
+Stream-Producer-Id: abc123
+Stream-Producer-Epoch: ?
+
+[]
+```
+
+Note: An empty body with `Stream-Producer-Epoch: ?` is a special case that only bumps the epoch and resets the sequence counter without writing data. The server responds with the new epoch:
```
kevin-dp: I dislike this special case. In 5.2 we explicitly prohibit an empty array as body:
> Servers MUST reject POST requests with an empty body (Content-Length: 0 or no body) with 400 Bad Request. Empty appends have no semantic meaning and are likely client errors.

But now we allow it in this very special case. I would rather accept any JSON body that isn't prohibited and document that it is ignored if `Stream-Producer-Epoch` is `?`. Any JSON body is fine but the most logical one to use would be `null`.

PROTOCOL.md
```diff
@@ -552,23 +746,29 @@ This port was selected from the IANA unassigned range 4434-4440. Standalone serv
...
+| `Stream-Producer-Id`    | permanent | This document |
+| `Stream-Producer-Epoch` | permanent | This document |
+| `Stream-Acked-Seq`      | permanent | This document |
 
 **Descriptions:**
 
 - `Stream-TTL`: Relative time-to-live for streams (seconds)
 - `Stream-Expires-At`: Absolute expiry time for streams (RFC 3339 timestamp)
-- `Stream-Seq`: Writer sequence number for coordination (opaque string)
+- `Stream-Seq`: Writer sequence number for coordination (opaque string or integer in idempotent mode)
```
kevin-dp: I'm a bit confused about this one being overloaded and having different semantics in regular vs idempotent mode. In regular mode it is used to coordinate writes such that in case of concurrent writes, only 1 will be able to write at that position in the stream. However, in idempotent mode this is used to detect duplicate writes.

**Does this mean that in idempotent mode we cannot coordinate writes using `Stream-Seq`?** That's a bit weird. Why not introduce a dedicated header for sequence number in idempotent mode such that we can still do the coordination.

KyleAMathews:
@kevin-dp I did a bunch more research and posted it in https://github.com/durable-streams/durable-streams/issues/108 — you want to port your thoughts over there as applicable? I'm going to close this PR and restart the work based on that issue.

I agree that a more producer-driven setup is simpler e.g. it produces the producer id, etc.

kevin-dp:
> @kevin-dp I did a bunch more research and posted it in #108 — you want to port your thoughts over there as applicable? I'm going to close this PR and restart the work based on that issue.
> 
> I agree that a more producer-driven setup is simpler e.g. it produces the producer id, etc.

@KyleAMathews I'm not entirely sure what the diff is between https://github.com/durable-streams/durable-streams/issues/108 and this PR. Please read through my comments on this PR and check what applies to https://github.com/durable-streams/durable-streams/issues/108

KyleAMathews:
@kevin-dp just posted a comment over there — closing this PR now

PR #65, opened 2025-12-17
feat: add an implementation of a DS spec as a caddy plugin

icehaunter:
Code is Claude-generated


KyleAMathews:
Great!

I think the two things to do are hook it up to CI to run the conformance tests & I know little about Go but some sort of release process.

This is a good candidate for the default dev DS server.

PR #68, opened 2025-12-18
Clarify offsets must be unique and strictly increasing

### LINKED ISSUE #63: Monotonic generation and collisions of offsets
Is the protocol fine if the offset used is a UTC timestamp for example, where collisions can happen, and monotonically increasing creation is not guaranteed? 

For example, if the offset is a UTC timestamp sent as a string to the server, what happens if the client goes "Resume from X offset", and the backing data storage has > 1 records with that offset? 

Likewise, is it required that the offset created on a new record be greater than all previously created offsets(I think the answer to this is yes, but it is not specified in the protocol description in the readme)?.

KyleAMathews:
Fix #63


jerebear12 (commented):
Looks great to me!

PR #70, opened 2025-12-19
state: synchronous createStreamDB.

thruflo:
The `createStreamDB` function in the state package doesn't need to be async and, if it's synchronous, this significantly improves the DX of working with it.

For example, in a `const { messages, client, collections } = useDurableChat(...)` hook, it avoids the client and collections ever being undefined which avoids a bunch of annoying conditional logic.


KyleAMathews (approved):
Ah nice, I blame Claude for the async part :shipit:

PR #101, opened 2025-12-24
Add TTL deletion tests to server conformance tests

KyleAMathews:
Add conformance tests that verify streams are properly deleted after their TTL expires or Expires-At time passes. Tests verify that HEAD, GET, and POST operations all return 404 after expiration.

Implement expiration checking in both in-memory and file-backed stores:
- Add isExpired() helper to check TTL and Expires-At
- Check expiration on all stream operations (get, has, read, append, etc.)
- Automatically delete expired streams when accessed

### DIFF
--- a/packages/caddy-plugin/store/file_store.go
+++ b/packages/caddy-plugin/store/file_store.go
@@ -103,12 +103,18 @@ func (s *FileStore) Create(path string, opts CreateOptions) (*StreamMetadata, bo
 	s.metaCacheMu.Lock()
 	defer s.metaCacheMu.Unlock()
 
-	// Check if stream already exists
+	// Check if stream already exists (and is not expired)
 	if existing, ok := s.metaCache[path]; ok {
-		if existing.ConfigMatches(opts) {
+		// If expired, delete it and allow recreation
+		if existing.IsExpired() {
+			if dirName, hasDirName := s.dirCache[path]; hasDirName {
+				s.deleteStreamUnlocked(path, dirName)
+			}
+		} else if existing.ConfigMatches(opts) {
 			return existing, false, nil
+		} else {
+			return nil, false, ErrConfigMismatch
 		}
-		return nil, false, ErrConfigMismatch
 	}
 
 	// Generate unique directory name
@@ -210,14 +216,18 @@ func (s *FileStore) Delete(path string) error {
 		return ErrStreamNotFound
 	}
 
+	s.deleteStreamUnlocked(path, dirName)
+	return nil
+}
+
+// deleteStreamUnlocked removes a stream without acquiring the lock (caller must hold metaCacheMu)
+func (s *FileStore) deleteStreamUnlocked(path string, dirName string) {
 	// Remove from writer pool
 	segPath := filepath.Join(s.dataDir, "streams", dirName, SegmentFileName)
 	s.writerPool.Remove(segPath)
 
-	// Delete from bbolt
-	if err := s.metaStore.Delete(path); err != nil {
-		return err
-	}
+	// Delete from bbolt (ignore errors on expired stream cleanup)
+	s.metaStore.Delete(path)
 
 	// Remove from cache
 	delete(s.metaCache, path)
@@ -228,8 +238,6 @@ func (s *FileStore) Delete(path string) error {
 	deletedDir := filepath.Join(s.dataDir, "streams", ".deleted~"+dirName+"~"+fmt.Sprintf("%d", time.Now().UnixNano()))
 	os.Rename(streamDir, deletedDir)
 	go os.RemoveAll(deletedDir)
-
-	return nil
 }
 
 // Append adds data to a stream
--- a/packages/caddy-plugin/store/memory_store.go
+++ b/packages/caddy-plugin/store/memory_store.go
@@ -41,13 +41,17 @@ func (s *MemoryStore) Create(path string, opts CreateOptions) (*StreamMetadata, 
 	s.mu.Lock()
 	defer s.mu.Unlock()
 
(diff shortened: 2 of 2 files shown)


samwillis:
Looks good - chatGPT is suggesting that we should also validate the aloud values in the expires-at and TTL headers:

---

## PR #101 review (protocol conformance focus)

### What this PR adds
- **6 new server conformance tests** that create a stream with either `Stream-TTL: 1` or `Stream-Expires-At: <now+1s>` and then assert:
  - it exists immediately (`HEAD`/`GET` ok), and
  - **after ~1.5s it’s treated as non-existent** (`HEAD`/`GET`/`POST` return `404`).  [oai_citation:0‡GitHub](https://github.com/durable-streams/durable-streams/pull/101/files)
- **File-backed store implementation** of expiry checking that lazily deletes expired streams on access via `getMetaIfNotExpired()` + `isExpired()` (checks `expiresAt` and `ttlSeconds`).  [oai_citation:1‡GitHub](https://github.com/durable-streams/durable-streams/pull/101/files)
- Changeset describes the intent: “streams return 404 after TTL/Expires-At passes” and that expired streams are deleted when accessed.  [oai_citation:2‡GitHub](https://github.com/durable-streams/durable-streams/pull/101/files)

### Does it validate conformance with `PROTOCOL.md`?
Yes — **for the specific behaviour “expired streams behave like they don’t exist”**, these tests line up with the spec:

- The spec defines durability as lasting “until the stream is deleted or expired”.  [oai_citation:3‡GitHub](https://raw.githubusercontent.com/durable-streams/durable-streams/refs/heads/main/PROTOCOL.md)  
- And it uses **`404 Not Found` for “stream does not exist”** on `POST` (append), `HEAD`, and `GET`.  [oai_citation:4‡GitHub](https://raw.githubusercontent.com/durable-streams/durable-streams/refs/heads/main/PROTOCOL.md)  
So asserting `404` after TTL/expiry passes is a good conformance check.

### Gaps vs the spec (things not covered yet)
These tests are narrowly scoped to “eventual 404 after expiry”. They **do not yet validate several MUST/SHOULD requirements in the TTL/expiry parts of the protocol**, for example:

1. **TTL syntax validation**: the spec requires `Stream-TTL` be a non-negative integer string with **no leading zeros, no + sign, no decimals/scientific notation**. No negative/invalid-header tests are added.  [oai_citation:5‡GitHub](https://raw.githubusercontent.com/durable-streams/durable-streams/refs/heads/main/PROTOCOL.md)  
2. **Mutual exclusion**: if both `Stream-TTL` and `Stream-Expires-At` are supplied, servers **SHOULD reject with `400 Bad Request`** (or document precedence). Not tested.  [oai_citation:6‡GitHub](https://raw.githubusercontent.com/durable-streams/durable-streams/refs/heads/main/PROTOCOL.md)  
3. **HEAD metadata semantics**: spec calls out `HEAD` as the canonical place to return TTL/expiry metadata (remaining TTL / absolute expiry). The new tests only check status codes, not the presence/meaning of those headers.  [oai_citation:7‡GitHub](https://raw.githubusercontent.com/durable-streams/durable-streams/refs/heads/main/PROTOCOL.md)  
4. **Idempotent PUT semantics with TTL/expiry**: spec says if stream exists, `PUT` must succeed with `200/204` only if configuration matches, else `409`. Not tested for TTL/expiry configs.  [oai_citation:8‡GitHub](https://raw.githubusercontent.com/durable-streams/durable-streams/refs/heads/main/PROTOCOL.md)  

### Implementation notes (server-side)
- The lazy “delete on access” approach in `file-store.ts` is consistent with the spec’s externally visible behaviour (client sees `404` after expiry), and avoids needing a background GC.  [oai_citation:9‡GitHub](https://github.com/durable-streams/durable-streams/pull/101/files)  
- **Potential TypeScript lint/compile issue:** in `file-store.ts`, several methods still declare `const key = ...` but no longer use it after switching to `getMetaIfNotExpired()`. If `noUnusedLocals`/lint is enabled, this can fail CI.  [oai_citation:10‡GitHub](https://github.com/durable-streams/durable-streams/pull/101/files)  
- I couldn’t view `packages/server/src/store.ts`’s diff in the rendered “Files changed” page from this environment (GitHub “failed to load files”); so I can’t independently confirm the in-memory store applies expiry checks to all operations the same way as the file-backed store.  [oai_citation:11‡GitHub](https://github.com/durable-streams/durable-streams/pull/101/files)  

### Suggestions to make conformance coverage “spec-complete” for TTL/expiry
If you want these to be true protocol conformance tests (not just regression tests), I’d add:
- Invalid `Stream-TTL` cases: `03600`, `+3600`, `-1`, `3600.0`, `3e3` → expect `400`.  [oai_citation:12‡GitHub](https://raw.githubusercontent.com/durable-streams/durable-streams/refs/heads/main/PROTOCOL.md)  
- `PUT` with **both** `Stream-TTL` and `Stream-Expires-At` → expect `400` (or assert your documented precedence rule).  [oai_citation:13‡GitHub](https://raw.githubusercontent.com/durable-streams/durable-streams/refs/heads/main/PROTOCOL.md)  
- `HEAD` returns TTL/expiry headers when set (and ideally that TTL decreases / expiry is stable), per spec.  [oai_citation:14‡GitHub](https://raw.githubusercontent.com/durable-streams/durable-streams/refs/heads/main/PROTOCOL.md)  
- Idempotent `PUT` behaviour when stream exists:
  - same config → `200/204`
  - different TTL/expiry → `409`.  [oai_citation:15‡GitHub](https://raw.githubusercontent.com/durable-streams/durable-streams/refs/heads/main/PROTOCOL.md)

PR #113, opened 2025-12-27
Add browser security headers

KyleAMathews:
Add Protocol Section 10.7 specifying browser security headers:
- X-Content-Type-Options: nosniff (prevents MIME sniffing)
- Cross-Origin-Resource-Policy: cross-origin (prevents CORB issues)
- Cache-Control: no-store on HEAD responses (per Section 5.4)

Update server implementation to set these headers on all responses.

Add conformance tests verifying security headers on:
- GET, PUT, POST, HEAD, SSE, long-poll responses
- Error responses (404, etc.)

### DIFF
--- a/packages/server/src/server.ts
+++ b/packages/server/src/server.ts
@@ -703,7 +703,7 @@ export class DurableStreamTestServer {
     // Track this SSE connection
     this.activeSSEResponses.add(res)
 
-    // Set SSE headers (including security headers since writeHead bypasses setHeader)
+    // Set SSE headers (explicitly including security headers for clarity)
     res.writeHead(200, {
       "content-type": `text/event-stream`,
       "cache-control": `no-cache`,


samwillis:
Review from ChatGPT:

------

Here’s my review of PR #113 (“Add browser security headers”). ([[GitHub](https://github.com/durable-streams/durable-streams/pull/113)][1])

## Overall

Solid, well-scoped change: you add a Protocol section (10.7) describing browser-focused security headers, update the TS server + Caddy plugin to emit them, and add conformance coverage for the common request modes (including SSE/long-poll + errors). ([[GitHub](https://github.com/durable-streams/durable-streams/pull/113/files)][2])

## What landed (as I see it)

* **PROTOCOL.md**: new section 10.7 documenting:

  * `X-Content-Type-Options: nosniff`
  * `Cross-Origin-Resource-Policy: cross-origin` (or same-origin/same-site)
  * `Cache-Control: no-store` on HEAD responses (+ optional `Content-Disposition`) ([[GitHub](https://github.com/durable-streams/durable-streams/pull/113/files)][2])
* **Caddy handler**: sets `X-Content-Type-Options` and `Cross-Origin-Resource-Policy` for responses (before preflight early-return). ([[GitHub](https://github.com/durable-streams/durable-streams/pull/113/files)][2])
* **Server (TS)**:

  * sets `x-content-type-options` + `cross-origin-resource-policy` early for responses
  * ensures SSE `writeHead` includes the security headers
  * adds `cache-control: no-store` on HEAD responses ([[GitHub](https://github.com/durable-streams/durable-streams/pull/113/files)][2])
* **Server conformance tests**: adds a “Browser Security Headers” suite validating `nosniff`, CORP presence/value, HEAD `no-store`, plus checks for SSE, long-poll, and 404s. ([[GitHub](https://github.com/durable-streams/durable-streams/pull/113/files)][2])
* **Changeset**: bumps `@durable-streams/server` + `@durable-streams/server-conformance-tests` as patch. ([[GitHub](https://github.com/durable-streams/durable-streams/pull/113/files)][2])

(Also: I could see that **caddy-conformance succeeded** in CI, but most logs require sign-in so I couldn’t inspect further.) ([[GitHub](https://github.com/durable-streams/durable-streams/actions/runs/20541284836/job/59005921955?pr=113)][3])

## Review notes / suggested changes

### 1) 🚨 GitHub is flagging “hidden or bidirectional Unicode text”

On the PR UI, GitHub shows the “hidden/bidirectional Unicode” warning on the changed files. That’s worth treating as **blocking** until confirmed/cleaned, because bidi control chars are a known supply-chain footgun.

Suggested quick scan locally:

```bash
python - <<'PY'
import pathlib, sys
# Bidi control ranges: https://trojansource.codes/
pat = ''.join([chr(c) for c in list(range(0x202A,0x202F)) + list(range(0x2066,0x206A))])
targets = [
  ".changeset/browser-security-headers.md",
  "PROTOCOL.md",
  "packages/server/src/server.ts",
  "packages/server-conformance-tests/src/index.ts",
  "packages/caddy-plugin/handler.go",
]
bad = False
for t in targets:
  p = pathlib.Path(t)
  if not p.exists(): 
    continue
  s = p.read_text(encoding="utf-8", errors="replace")
  for i,ch in enumerate(s):
    o = ord(ch)
    if 0x202A <= o <= 0x202E or 0x2066 <= o <= 0x2069:
      print(f"{t}: contains bidi control U+{o:04X} at index {i}")
      bad = True
if bad:
  sys.exit(1)
print("No bidi control chars found in target files.")
PY
```

### 2) Spec language vs conformance strictness

Protocol 10.7 uses **SHOULD** wording, but the conformance suite effectively makes at least `nosniff` and HEAD `no-store` *required* (for implementations running these conformance tests). ([[GitHub](https://github.com/durable-streams/durable-streams/pull/113/files)][2])
That might be fine, but I’d suggest one of:

* tighten the spec language for what conformance expects (“MUST include … for browser-facing servers”), **or**
* loosen conformance to treat these as recommended (not failing) if you truly want SHOULD.

### 3) Server implementation: “writeHead bypasses setHeader” consistency

You handled SSE correctly by explicitly including the security headers in the `writeHead(200, { ... })` call. ([[GitHub](https://github.com/durable-streams/durable-streams/pull/113/files)][2])
Given that, I’d consider doing the same for the HEAD `writeHead` path too (or factor a `commonHeaders()` helper that returns a header object). Right now, HEAD relies on Node’s merge semantics between `setHeader` and `writeHead`, which is easy to accidentally break later.

### 4) Conformance tests: reduce “abort means we didn’t assert” gaps

The SSE + long-poll tests accept `AbortError` and then don’t assert headers in that case. ([[GitHub](https://github.com/durable-streams/durable-streams/pull/113/files)][2])
If CI/network timing ever shifts, you can end up with a “pass” that didn’t actually validate anything. A more deterministic pattern is:

* start SSE / long-poll,
* ensure you receive headers (or a first byte),
* then abort/close.

Also, there’s an autofix/lint annotation about an `async` callback with no `await`—worth cleaning for signal/noise. ([[GitHub](https://github.com/durable-streams/durable-streams/pull/113/files)][2])

### 5) Minor: broaden CORP coverage (optional)

You only assert CORP on GET (and `nosniff` on the error case). ([[GitHub](https://github.com/durable-streams/durable-streams/pull/113/files)][2])
Given the spec text says “on all responses,” you might optionally add:

* CORP assertion on at least one non-GET response (PUT/POST),
* CORP on 404 responses too.

## Verdict

**Looks good conceptually and the direction is right.** If you:

1. clear the bidi/hidden-unicode warning, and
2. tighten up the couple of test/implementation consistency points above,

…I’d be happy with this landing.

[1]: https://github.com/durable-streams/durable-streams/pull/113 "Add browser security headers by KyleAMathews · Pull Request #113 · durable-streams/durable-streams · GitHub"
[2]: https://github.com/durable-streams/durable-streams/pull/113/files "Add browser security headers by KyleAMathews · Pull Request #113 · durable-streams/durable-streams · GitHub"
[3]: https://github.com/durable-streams/durable-streams/actions/runs/20541284836/job/59005921955?pr=113 "Add browser security headers · durable-streams/durable-streams@78031ea · GitHub"

PR #115, opened 2025-12-27
Extract Server Routes into Standalone DurableStreamRouter Class

lordofthelake:
## Summary

This PR extracts the HTTP route handling logic from `DurableStreamTestServer` into a standalone `DurableStreamRouter` class that can be embedded into any Node.js HTTP server (Express, Fastify, Koa, vanilla Node.js, etc.).

**Key Benefits:**
- ✅ Router can be used as a standalone component in existing servers
- ✅ Fault injection isolated into separate test-only middleware
- ✅ Zero breaking changes - all existing tests pass
- ✅ Optional CORS headers (configurable)
- ✅ Automatic path prefix stripping for sub-route mounting
- ✅ Cleaner separation of concerns

## Motivation

The existing `DurableStreamTestServer` implementation tightly coupled route handling with server lifecycle management, making it impossible to embed the stream API into existing HTTP servers. This prevented users from:

- Adding durable streams as a sub-route in existing applications (e.g., `/api/streams/*`)
- Using custom authentication/authorization middleware
- Integrating with existing Express/Fastify applications
- Composing with other HTTP middleware

### DIFF
--- a/packages/server/src/registry-hook.ts
+++ b/packages/server/src/registry-hook.ts
@@ -3,19 +3,18 @@
  * This stream records all create/delete events for observability.
  */
 
-import { DurableStream } from "@durable-streams/client"
-import { createStateSchema } from "@durable-streams/state"
-import type { StreamLifecycleHook } from "./types"
-import type { StreamStore } from "./store"
-import type { FileBackedStreamStore } from "./file-store"
+import { DurableStream } from "@durable-streams/client";
+import { createStateSchema } from "@durable-streams/state";
+import type { StreamLifecycleHook } from "./types";
+import type { StreamStorage } from "./store";
 
-const REGISTRY_PATH = `/v1/stream/__registry__`
+const REGISTRY_PATH = `/v1/stream/__registry__`;
 
 // Registry schema for the server
 interface StreamMetadata {
-  path: string
-  contentType: string
-  createdAt: number
+  path: string;
+  contentType: string;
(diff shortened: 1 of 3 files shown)


lordofthelake:
> Yeah bun, demo, and cloudflare workers. The latter also now supports node:http so perhaps that could be sufficient. Ideally we'd just use web request/response objects and have a node:http wrapper as that's the most portable. Though if that's too hard (I'm not sure!) it's not critical as this is just a reference implementation.

@KyleAMathews I can make the Router implementation more minimal and just depend on `node:http`, and split everything else (CORS, compression) into middlewares, so that the core handler has a very minimal surface that is as  portable as possible. For the HTTP req/res the other server-side "standard", apart from the Node.js one is the WhatCG one, but that would require an adapter for Node or adding a dependency on something like Hono (that has several adapters). What's your preference?

KyleAMathews:
> What would be your preference?

Well, tbh I'm having second thoughts about this refactor. The goal of the node server isn't to serve as a base for production but as a "reference implementation" e.g. all of our tests run against it etc. I haven't used express seriously in years and don't really remember how it works and have never used Bun.serve.

So perhaps a better approach is an express ready code base just goes in a different repo. And a Bun.serve goes in a different repo. With the conformance tests and e.g. Claude code you could have something running fairly quickly. Just checkout this repo alongside your new one and have the agent run away with it.

Sound good?

PR #123, opened 2025-12-30
fix: use bundled vitest config for conformance tests CLI

balegas:
## Summary

- Fixes issue where `npx @durable-streams/server-conformance-tests --run <url>` would fail when run from a directory containing a local `vitest.config.ts`
- Vitest was picking up the local config instead of running the conformance tests standalone

## Changes

- Added `vitest.conformance.config.ts` bundled with the package
- Updated CLI to pass `--config` flag pointing to the bundled config
- Updated `eslint.config.js` to ignore all vitest config files (`**/vitest*.config.ts`)

## Test plan

- [x] Run `node packages/server-conformance-tests/dist/cli.js --run <url>` from monorepo root (previously failed, now works)
- [ ] Publish and test with `npx @durable-streams/server-conformance-tests --run <url>` from an external project


balegas:
I couldn't run them with `npx`. @KyleAMathews did you try it?

PR #126, opened 2025-12-31
Warn in the client when using HTTP for stream URLs.

thruflo:
Add a console warning when using HTTP (not HTTPS) URLs in browser environments. HTTP limits browsers to 6 concurrent connections per host (HTTP/1.1), which can cause slow streams and app freezes when multiple streams are active.

The warning can be disabled by setting warnOnHttp: false in options. It is automatically suppressed in Node.js environments and during tests.

Note, this mirrors the implementation here: https://github.com/electric-sql/electric/pull/3661


thruflo:
@KyleAMathews addressed (and read code and tested in browser).

KyleAMathews (approved):

PR #133, opened 2026-01-05
Add support for offset?=now

### LINKED ISSUE #1: Add support for 'offset=now' to skip historical data

balegas:
Trusting the agents. Haven't experimented it yet.

Closes #1

### DIFF
--- a/PROTOCOL.md
+++ b/PROTOCOL.md
@@ -514,7 +514,7 @@ This enables CDN/proxy caching while allowing stale content to be served during 
 
 **ETag Usage:**
 
-Servers **MUST** generate `ETag` headers for GET responses. Clients **MAY** use `If-None-Match` with the `ETag` value on repeat catch-up requests. When a client provides a valid `If-None-Match` header that matches the current ETag, servers **MUST** respond with `304 Not Modified` (with no body) instead of re-sending the same data. This is essential for fast loading and efficient bandwidth usage.
+Servers **MUST** generate `ETag` headers for GET responses, except for `offset=now` responses. Clients **MAY** use `If-None-Match` with the `ETag` value on repeat catch-up requests. When a client provides a valid `If-None-Match` header that matches the current ETag, servers **MUST** respond with `304 Not Modified` (with no body) instead of re-sending the same data. This is essential for fast loading and efficient bandwidth usage.
 
 **Query Parameter Ordering:**
 


PROTOCOL.md
```diff
@@ -397,7 +397,17 @@ Offsets are opaque tokens that identify positions within a stream. They have the
...
 
-**Sentinel Value**: The special offset value `-1` represents the beginning of the stream. Clients **MAY** use `offset=-1` as an explicit way to request data from the start. This is semantically equivalent to omitting the offset parameter. Servers **MUST** recognize `-1` as a valid offset that returns data from the beginning of the stream.
+**Sentinel Values**: The protocol defines two special offset sentinel values:
+
+- **`-1` (Stream Beginning)**: The special offset value `-1` represents the beginning of the stream. Clients **MAY** use `offset=-1` as an explicit way to request data from the start. This is semantically equivalent to omitting the offset parameter. Servers **MUST** recognize `-1` as a valid offset that returns data from the beginning of the stream.
+
+- **`now` (Current Tail Position)**: The special offset value `now` allows clients to skip all existing data and begin reading from the current tail position. This is useful for applications that only care about future data (e.g., presence tracking, live monitoring, late joiners to a conversation). When a client requests `offset=now`:
+  - Servers **MUST** return `200 OK` with an empty response body
```
samwillis: I would be inclined to special case json mode to return `[]`

samwillis:
Review from ChatGPT - https://chatgpt.com/share/695e708f-3f0c-8005-a9b6-5c8b6d136f1c

----


### High-level: does PR #133 solve issue #1?

Yes — it implements the exact “jump to tail” capability issue #1 asks for: a special sentinel `offset=now` that lets clients start following *from the current end* without replaying history (presence, monitoring, late joiners), and it explicitly supports combining with live modes (long-poll/SSE) to avoid the “catch up first” requirement. ([[GitHub](https://github.com/durable-streams/durable-streams/issues/1)][1])

### What’s good / aligned with the issue + protocol shape

* **Correct core semantics for `offset=now`**:

  * Catch-up: empty response + tail in `Stream-Next-Offset` + `Stream-Up-To-Date: true`. ([[GitHub](https://github.com/durable-streams/durable-streams/issues/1)][1])
  * Long-poll: avoids a round-trip by *waiting immediately* (timeout → `204 No Content` with up-to-date true). ([[GitHub](https://github.com/durable-streams/durable-streams/pull/133/commits/55b7efd94c616e8f1f7ea21597c02d47715bb82c)][2])
  * SSE: converts `now` to “tail offset” and streams future events only. ([[GitHub](https://github.com/durable-streams/durable-streams/pull/133/commits/55b7efd94c616e8f1f7ea21597c02d47715bb82c)][2])

* **Encoding-aware empty body**: the Caddy path returns `[]` for JSON streams and an empty body otherwise, and the TS server path does the same (as documented). That’s important because “empty but valid JSON” matters for strict clients. ([[GitHub](https://github.com/durable-streams/durable-streams/pull/133/commits/e49567be981e832cf36aa07d5d74d31c183694be)][3])

* **Caddy plugin implementation is clean and safe**:

  * `ParseOffset` allows `"now"` explicitly and still keeps strict validation for real offsets. ([[GitHub](https://github.com/durable-streams/durable-streams/pull/133/commits/e49567be981e832cf36aa07d5d74d31c183694be)][3])
  * SSE path converts `now` → `meta.CurrentOffset` before entering the SSE handler (so internal code keeps working with real offsets). ([[GitHub](https://github.com/durable-streams/durable-streams/pull/133/commits/e49567be981e832cf36aa07d5d74d31c183694be)][3])
  * Catch-up response sets the required headers and no payload. ([[GitHub](https://github.com/durable-streams/durable-streams/pull/133/commits/e49567be981e832cf36aa07d5d74d31c183694be)][3])
  * Follow-up fix ensures **`Stream-Cursor` is present even on 204 long-poll timeouts**, which is easy to miss and good for spec compliance. ([[GitHub](https://github.com/durable-streams/durable-streams/pull/133/commits/02a67208ee13d02473d56f138bee755258de3361)][4])

* **Spec hygiene**: reserving `-1` and `now` so implementations never emit them as real offsets is the right move (prevents ambiguity and weird edge cases). ([[GitHub](https://github.com/durable-streams/durable-streams/pull/133/commits/4561341da7a76cd9a3312a3d6718b08bb292b0bf)][5])

### Things I would double-check before merging

1. **TS server long-poll “wait” call uses the effective offset everywhere**

   * In the TS server, `offset=now` gets converted to an `effectiveOffset` (tail) and reads use that — good. ([[GitHub](https://github.com/durable-streams/durable-streams/pull/133/commits/55b7efd94c616e8f1f7ea21597c02d47715bb82c)][2])
   * I’d quickly sanity-check that the value passed into the *wait logic* is consistently `effectiveOffset` rather than the literal `'now'` sentinel (depends on how `waitForMessages(...)` interprets its arguments). The intent is correct (“treat as caught up and wait”), but this is the one place a sentinel can accidentally leak into deeper store logic.

2. **ETag + Cache-Control interplay**

   * You’re generating ETags for `offset=now` responses and later added `Cache-Control: no-store` for those catch-up responses (per commits list). That’s fine, but:

     * consider whether ETag is useful at all when you explicitly don’t want storage/collapsing for “now” positioning responses (some CDNs behave unexpectedly unless you’re explicit).
     * make sure TS and Go shapes are consistent if you rely on them in conformance.

3. **Missing explicit conformance coverage for “empty stream” edge case**

   * Issue #1’s checklist explicitly calls out `offset=now` on an *empty* stream. ([[GitHub](https://github.com/durable-streams/durable-streams/issues/1)][1])
   * I saw long-poll timeout coverage and “append after poll” coverage in the conformance tests. ([[GitHub](https://github.com/durable-streams/durable-streams/pull/133/commits/55b7efd94c616e8f1f7ea21597c02d47715bb82c)][2])
   * If there isn’t already a test that:

     * creates an empty stream, calls `GET ?offset=now` and asserts `Stream-Next-Offset` is valid (whatever “zero” means in your offset space), and body is empty/`[]`,
     * then follows with long-poll / SSE,
       …I’d add it. It’s the most likely place for an off-by-one or “missing meta” bug.

4. **404 / non-existent stream semantics**

   * Worth verifying: `offset=now` should not accidentally “create” a stream or mask a missing stream (especially in implementations where “meta” fetch is lazy). A conformance test here would be cheap and would prevent regressions.

### Nits / small polish suggestions

* In the Go plugin, the `NowOffset` sentinel is implemented as max uint64 values and documented as “guaranteed never to be valid offsets” — good. ([[GitHub](https://github.com/durable-streams/durable-streams/pull/133/commits/4561341da7a76cd9a3312a3d6718b08bb292b0bf)][5])
  I’d just ensure no codepath ever serialises that sentinel via `Offset.String()` into a response header (your current logic seems to avoid that by using `meta.CurrentOffset`). ([[GitHub](https://github.com/durable-streams/durable-streams/pull/133/commits/e49567be981e832cf36aa07d5d74d31c183694be)][3])

* Python typing change (`data: T | None`) makes sense for “control-only” events when using `offset=now` + SSE and no data has arrived yet. ([[GitHub](https://github.com/durable-streams/durable-streams/pull/133/commits/4561341da7a76cd9a3312a3d6718b08bb292b0bf)][5])

### Bottom line

This PR is on the right track and looks like it implements the intended UX from issue #1 very directly (including the “no extra round-trip” optimisation for long-poll/SSE), with good spec updates and conformance tightening. ([[GitHub](https://github.com/durable-streams/durable-streams/issues/1)][1])

If you want, paste the TS `waitForMessages` signature/body (or point me at the exact lines) and I’ll give you a very concrete “yes, sentinel can’t leak / here’s the one-line fix” check.

[1]: https://github.com/durable-streams/durable-streams/issues/1 "Add support for 'offset=now' to skip historical data · Issue #1 · durable-streams/durable-streams · GitHub"
[2]: https://github.com/durable-streams/durable-streams/pull/133/commits/55b7efd94c616e8f1f7ea21597c02d47715bb82c "Add support for offset?=now by balegas · Pull Request #133 · durable-streams/durable-streams · GitHub"
[3]: https://github.com/durable-streams/durable-streams/pull/133/commits/e49567be981e832cf36aa07d5d74d31c183694be "Add support for offset?=now by balegas · Pull Request #133 · durable-streams/durable-streams · GitHub"
[4]: https://github.com/durable-streams/durable-streams/pull/133/commits/02a67208ee13d02473d56f138bee755258de3361 "Add support for offset?=now by balegas · Pull Request #133 · durable-streams/durable-streams · GitHub"
[5]: https://github.com/durable-streams/durable-streams/pull/133/commits/4561341da7a76cd9a3312a3d6718b08bb292b0bf "Add support for offset?=now by balegas · Pull Request #133 · durable-streams/durable-streams · GitHub"

balegas:
## Addressed Review Feedback

Thanks for the thorough reviews! Here's how we addressed the feedback:

### 🔴 Critical Bug Fixed
- **`_afetch_next` → `_fetch_next`**: Fixed the incorrect method name in `conformance_adapter_async.py` (ef6d5ed)

### 🟡 ETag + Cache-Control Interplay
- **Removed unnecessary ETag from `offset=now` responses**: Since we're setting `Cache-Control: no-store`, the ETag is unnecessary and could cause unexpected CDN behavior. Removed from both TS server and Go/Caddy plugin (5fd72b1)

### 🟡 Missing Conformance Coverage

Added tests for empty stream edge cases and 404 semantics:

**Client Conformance Tests** (bec48d6):
- `offset-now-empty-stream-longpoll` - offset=now on empty stream with long-poll
- `offset-now-empty-stream-sse` - offset=now on empty stream with SSE  
- `read-offset-now-nonexistent` - offset=now on non-existent stream returns 404
- `read-offset-now-nonexistent-longpoll` - long-poll with offset=now on non-existent stream
- `read-offset-now-nonexistent-sse` - SSE with offset=now on non-existent stream

**Server Conformance Tests** (21799a4):
- `should return 404 for offset=now with long-poll on non-existent stream`
- `should return 404 for offset=now with SSE on non-existent stream`
- `should support offset=now with long-poll on empty stream` - verifies 204 timeout and valid resumable offset
- `should support offset=now with SSE on empty stream` - verifies upToDate:true with valid offset

### ✅ Python Typing Change
- The `data: T | None` change is already documented in the `StreamEvent` docstring explaining it may be `None` for control-only events (e.g., when using `offset=now` with SSE and no data has arrived yet)

samwillis (approved):
LGTM - just the etag in the protocol suggestion from @KyleAMathews?

PR #135, opened 2026-01-05
feat(state): add per-collection field parsers for JSON transformation

samwillis:
## Summary

Add per-collection field parsers for transforming raw JSON values during event dispatch. This follows Electric's parser pattern but operates per-collection and per-field since there's no database schema.

## Changes

- Add `FieldParsers<T>` type mapping field names to transform functions
- Add optional `parser` property to `CollectionDefinition`
- Update `EventDispatcher` to store and apply parsers per event type
- Register parsers in `createStreamDB` when setting up collections
- Add tests for date parsing, missing fields, and multiple parsers

## Usage Example

```typescript
const stateSchema = createStateSchema({
  events: {
    schema: eventSchema,
    type: 'event',
    primaryKey: 'id',
    parser: {
      createdAt: (v: unknown) => new Date(v as string),
      updatedAt: (v: unknown) => new Date(v as string),
    },
  },
})
```

## Testing

All 51 tests pass, including 3 new tests:
- `should apply field parsers to transform incoming stream data`
- `should handle parsers for fields not present in event`
- `should apply multiple parsers on same collection`

---

> [!NOTE]
> Adds field-level JSON-to-type transformation during event dispatch.
> 
> - New `FieldParsers<T>` type and optional `parser` on `CollectionDefinition` for per-field transforms
> - `EventDispatcher` stores parsers per event type and applies them to `event.value` before setting the primary key and writing
> - `createStreamDB` registers collection-defined parsers with the dispatcher
> - Tests cover date parsing, optional/missing fields, and multiple parsers in one collection
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 02067d7910a8d892151ee33b7b8f1018ee4a228c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/state/src/stream-db.ts
+++ b/packages/state/src/stream-db.ts
@@ -240,7 +240,10 @@ class EventDispatcher {
   private handlers = new Map<string, CollectionSyncHandler>()
 
   /** Map from event type to field parsers */
-  private parsers = new Map<string, FieldParsers<unknown>>()
+  private parsers = new Map<
+    string,
+    Record<string, (value: unknown) => unknown>
+  >()
 
   /** Handlers that have pending writes (need commit) */
   private pendingHandlers = new Set<CollectionSyncHandler>()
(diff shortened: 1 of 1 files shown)


KyleAMathews:
@balegas yeah interesting point — this isn't a parsing problem like we had with Electric — @samwillis what use case drove you to add this?

PR #140, opened 2026-01-05
Implement Kafka-style idempotent producers

### LINKED ISSUE #108: Exactly-Once Write Primitives: Research & Recommendation
Below is a research document I put together with Claude Code reviewing our current primitives & proposals for ensuring exactly-once writes.

It conforms that Kafka's idempotent producer approach w/ pipelining is now standard & we should just default to it & remove the append API.

---

## Executive Summary

This document analyzes approaches for ensuring fast, exactly-once writes from producers in durable-streams. After surveying Kafka, Pulsar, RabbitMQ Streams, NATS JetStream, Redis Streams, AWS Kinesis, Google Pub/Sub, Azure Event Hubs, and academic research, we recommend:

**Primary primitive**: Kafka-style idempotent producer with pipelining
- Simple API: fire writes, await flush, handle errors via callback
- Fast: 5 in-flight batches eliminates RTT bottleneck (5x current throughput)
- Safe: server-side deduplication via producer ID + sequence numbers
- Proven: 8+ years in production at Kafka, adopted by Pulsar and RabbitMQ

**Secondary (optional)**: OCC via `If-Match` header for multi-writer coordination

**Defer**: WebSocket API (optimization, not semantic guarantee)

---

## Table of Contents

1. [Theoretical Foundation](#theoretical-foundation)
2. [Industry Survey](#industry-survey)
3. [Current State in durable-streams](#current-state-in-durable-streams)
4. [Analysis of Proposed Features](#analysis-of-proposed-features)
5. [Recommendation](#recommendation)
6. [Proposed API](#proposed-api)
7. [Implementation Details](#implementation-details)
8. [Migration Path](#migration-path)
9. [References](#references)

---

### The Impossibility Result

Exactly-once *delivery* is mathematically impossible. This follows from two foundational results:

1. **Two Generals Problem**: Consensus over a lossy network cannot be guaranteed. If a protocol existed for exactly-once delivery, it could solve the Two Generals Problem—but we know that's impossible.

2. **FLP Impossibility** (Fischer, Lynch, Paterson 1985): No algorithm deterministically solves consensus in an asynchronous environment when even one process can crash.

### What We Can Achieve

While exactly-once *delivery* is impossible, exactly-once *processing* (or "effectively-once semantics") is achievable through:

```
Exactly-Once Processing = At-Least-Once Delivery + Idempotent Deduplication
```

As Jay Kreps (Kafka co-creator) clarified:
> "'Delivery' is a transport semantic. 'Processing' is an application semantic."

The practical approach used by all major streaming systems:
1. Retry failed sends (at-least-once delivery)
2. Deduplicate at the receiver (idempotent processing)
3. Result: each message affects state exactly once

---

### Comparison Matrix

| System | Dedup Mechanism | Dedup Scope | Pipelining | Producer Restart | Overhead |
|--------|-----------------|-------------|------------|------------------|----------|
| **Kafka** | Producer ID + Epoch + Seq | Unbounded (in log) | 5 in-flight | Epoch bump | Very low |
| **Pulsar** | Producer Name + Seq ID | Configurable | Yes | Query last seq | Very low |
| **RabbitMQ Streams** | Producer Name + Publish ID | Unbounded | Yes | Query last ID | Low |
| **NATS JetStream** | `Nats-Msg-Id` header | 2 min window | Yes | App manages | Low |
| **Redis Streams** | Consumer-side (PEL) | N/A | N/A | N/A | Medium |
| **AWS Kinesis** | None built-in | N/A | N/A | App manages | N/A |
| **Google Pub/Sub** | Server-side (pull only) | Session-based | N/A | N/A | Medium |
| **Azure Event Hubs** | Seq numbers (optional) | Session-based | Limited | Query state | Medium |

#### Apache Kafka (KIP-98)

The industry gold standard, introduced in Kafka 0.11 (2017), default since Kafka 3.0.

**Mechanism:**
- **Producer ID (PID)**: Assigned by broker on initialization
- **Epoch**: Incremented on producer restart, prevents "zombie" producers
- **Sequence number**: Per topic-partition, monotonically increasing per batch
- **Broker validation**:
  - `seq < last`: Duplicate, return success (idempotent)
  - `seq == last + 1`: Accept, update last
  - `seq > last + 1`: Reject (gap indicates lost messages)

**Key design decisions:**
- Sequence state persisted *in the log itself* (no separate state store)
- Pipelining up to 5 in-flight requests without breaking ordering
- Overhead: "just a few extra numeric fields with each batch"

**Performance:** "For the same value of max.in.flight.requests and acks, enabling idempotence alone does not significantly impact performance."

#### Apache Pulsar (PIP-6)

Similar to Kafka, with explicit producer naming.

**Mechanism:**
- **Producer Name**: User-provided, stable identifier
- **Sequence ID**: Application-managed, strictly increasing
- **Deduplication cursor**: Stored in BookKeeper
- **Broker maintains**: Map of Producer Name → Last Sequence ID

**Key difference:** Producer name is explicit (good for debugging), not auto-assigned.

**Performance:** "Extensive testing has found no measurable overhead in terms of throughput or latency degradation when deduplication is enabled."

#### RabbitMQ Streams

Follows the Kafka pattern closely.

**Mechanism:**
- **Producer Name**: Must be unique per stream, stable across restarts
- **Publishing ID**: Strictly increasing (gaps allowed)
- **Broker filtering**: Rejects ID ≤ last confirmed, still sends ack
- **Recovery**: `GetLastPublishedId()` API for resuming after restart

**Key insight:** Even filtered duplicates receive confirmations—the producer doesn't need to know if it was a duplicate.

#### NATS JetStream

Time-windowed deduplication with a header-based approach.

**Mechanism:**
- **Message header**: `Nats-Msg-Id` for deduplication
- **Time window**: 2 minutes default (configurable in nanoseconds)
- **Double ack**: Server acknowledges the client's ack

**Alternative approach:** `DiscardNewPerSubject` with max 1 message per subject provides infinite deduplication by embedding the unique ID in the subject name.

**Trade-off:** Simpler implementation, but time-bounded window may miss duplicates after window expires.

#### Redis Streams

Different model—consumer-side deduplication only.

**Mechanism:**
- **No producer dedup**: Server assigns message IDs
- **Pending Entries List (PEL)**: Tracks delivered-but-unacked messages
- **XACK**: Explicit acknowledgment removes from PEL
- **Message claiming**: Other consumers can claim stuck messages

**Trade-off:** Shifts deduplication burden to consumers. Not suitable when producer-side exactly-once is required.

#### AWS Kinesis

No native producer-side deduplication.

**Recommendation from AWS:**
- Embed idempotency keys in records
- Deduplicate at consumer using DynamoDB conditional writes
- Design consumers to be naturally idempotent

**Quote:** "Consumer applications should be designed to handle duplicate records gracefully since Kinesis guarantees at-least-once delivery."

#### Google Cloud Pub/Sub

Server-side deduplication for pull subscriptions (GA December 2022).

**Limitations:**
- Pull subscriptions only (not push)
- Single-region guarantee only
- ~1000 msg/sec throughput with ordering keys

#### Azure Event Hubs

Optional idempotent publishing (design phase).

**Mechanism:**
- Producer Group ID + Owner Level + Sequence Number
- Client queries partition state on connect

**Important caveat:** "Enabling idempotent retries does not guarantee exactly-once semantics. The existing Event Hubs at-least-once delivery contract still applies."

#### MillWheel (Google, 2013)

Seminal paper on exactly-once stream processing at scale.

**Key concepts:**
- **Strong productions**: Checkpoint before sending downstream
- **Weak productions**: Opt-out for performance-critical paths
- **Low watermarks**: Logical time for aggregations

**Quote:** "Billing pipelines depend on MillWheel's exactly-once guarantees."

#### Chandy-Lamport Snapshots (1985)

Foundation for Flink's checkpointing:
- Barrier injection into data streams
- Distributed consistent snapshots
- Recovery = restore state + replay from source

#### State Machine Replication

Zab (ZooKeeper) enforces idempotent state transitions—same state change applied multiple times yields the same result.

---

### What We Have

| Feature | Implementation | Guarantee |
|---------|----------------|-----------|
| **Sequence numbers** | `Stream-Seq` header, lexicographic comparison | Writer coordination (prevents interleaving) |
| **Batching** | `fastq` with concurrency=1 | One POST in flight, messages accumulated |
| **Auto-batching** | Enabled by default | Opportunistic batching while request in flight |

### What's Missing

1. **No producer identity**: Can't distinguish retry from new write
2. **No server-side deduplication**: Timeout + retry = duplicate
3. **Limited throughput**: Concurrency=1 means one RTT per batch

### The Gap

Current sequence numbers prevent *interleaving* from concurrent writers, but don't prevent *duplication* from retries:

```
Producer sends batch (seq=5)
    → Network timeout (did server receive it?)
Producer retries with seq=6
    → If server received both: duplicate data
```

---

### 1. Kafka-style Idempotent Producer (PR #51)

**Problem solved:** "Network failures create uncertainty—when a request times out, you don't know if the server received it."

**Proposed mechanism:**
- Producer ID + Epoch assigned by server
- Sequence numbers per batch
- Server deduplicates: returns 204 for duplicates
- Pipelining: up to 5 in-flight batches

**Verdict: ✅ RECOMMENDED as core primitive**

This is the industry-proven pattern used by Kafka, Pulsar, and RabbitMQ. It directly addresses the gap in our current implementation.

### 2. OCC via If-Match (Issue #32)

**Problem solved:** "Multiple writers need to coordinate—detect if someone else wrote since I last read."

**Proposed mechanism:**
- `If-Match` header with expected offset
- 412 Precondition Failed on mismatch
- `Stream-Next-Offset` in response for retry logic

**Verdict: ✅ KEEP as optional feature (different use case)**

OCC and idempotent producers solve *different* problems:

| Scenario | Idempotent Producer | OCC |
|----------|---------------------|-----|
| Retry after timeout | ✅ Handles safely | ❌ May get 412 |
| Concurrent writers | ❌ Both succeed | ✅ Conflict detection |
| "Append only if offset=X" | ❌ Not supported | ✅ Core feature |

**Example where OCC matters but idempotent producer doesn't help:**
```
Writer A reads stream at offset 5, prepares update
Writer B appends, stream now at offset 6
Writer A appends:
  - With idempotent producer: succeeds (no duplicate)
  - With OCC (If-Match: 5): returns 412 (conflict detected)
```

### 3. WebSocket API (Issue #5)

**Problem solved:** "HTTP request/response adds latency; WebSocket enables continuous streaming."

**Proposed mechanism:**
- Persistent connection
- Async acknowledgments after fsync
- Sliding window for in-flight messages

**Verdict: ⏸️ DEFER**

WebSocket is a *transport optimization*, not a *semantic guarantee*. Analysis:

1. **HTTP + pipelining already fast**: 5 in-flight batches eliminates most RTT overhead
2. **Complexity cost**: Reconnection logic, state synchronization, proxy issues
3. **Industry precedent**: Kafka and Pulsar use TCP/HTTP, not WebSocket
4. **Can add later**: Clear optimization target if latency becomes critical

---

### Primary: Kafka-style Idempotent Producer

Implement the Kafka-style idempotent producer as the core exactly-once primitive.

**Why:**
1. **Industry proven**: 8+ years in Kafka, adopted by Pulsar, RabbitMQ
2. **Simple API**: Fire-and-forget with error callback
3. **Fast**: 5x throughput improvement via pipelining
4. **Low overhead**: Few numeric fields per batch
5. **Matches use case**: Single producer streaming data

### Secondary: OCC (Optional)

Keep OCC (`If-Match` header) as an optional feature for multi-writer coordination.

**Why:**
- Different use case from idempotent producer
- Some users genuinely need compare-and-swap semantics
- Can be used alongside idempotent producer

### Defer: WebSocket

Defer WebSocket API until latency is a proven bottleneck.

**Why:**
- Pipelining provides most of the benefit
- Significant implementation complexity
- No semantic improvement, only performance

### Relationship to Current Primitives

| Primitive | Keep? | Relationship |
|-----------|-------|--------------|
| `Stream-Seq` header | Yes | App-level coordination, complementary |
| Batching (concurrency=1) | Replace | Upgrade to pipelining (concurrency=5) |
| `If-Match` (OCC) | Yes | Optional, orthogonal feature |

### Two-Layer Sequence Design

The system uses **two independent sequence mechanisms** that serve different purposes:

| Layer | Headers | Scope | Purpose | Managed by |
|-------|---------|-------|---------|------------|
| **Transport dedup** | `X-Producer-Id` + `X-Producer-Seq` | Per-producer session | Retry safety within session | Client library (automatic) |
| **Application coord** | `Stream-Seq` | Per-stream (all producers) | Cross-restart/cross-producer ordering | Application (opt-in) |

**Why both are needed:**

#### Transport-level deduplication (automatic)
```
Producer session: writes seq 0, 1, 2
Network timeout on seq 2
Retry seq 2 → server sees duplicate, returns 204
```
Handled automatically by `IdempotentProducer`. Users don't think about it.

#### Application-level coordination (opt-in)

**Scenario 1: Cross-restart protection**
```
Session 1: writes to offsets 0-99, crashes
Session 2: restarts, doesn't know where it left off
           idempotent producer resets to seq=0, epoch bumps
           → could accidentally re-send same data

With Stream-Seq:
           application tracks "I wrote up to Stream-Seq 99"
           new session sends Stream-Seq: 99
           → server rejects if stream already past that point
```

**Scenario 2: Multi-producer coordination**
```
Producer A (order-service-1): writes to "events" stream
Producer B (order-service-2): also writes to "events" stream

X-Producer-Seq: each has independent sequence, no coordination
Stream-Seq: both can coordinate against the same stream-level sequence
```

**Scenario 3: Exactly-once across system boundaries**
```
External system: "process events starting at offset 500"
Producer sends Stream-Seq: 499
→ server rejects if accidentally re-sending earlier events
```

**Combined usage:**
```http
POST /stream/my-stream HTTP/1.1
X-Producer-Id: order-service-1    # Transport dedup (automatic)
X-Producer-Seq: 42                # This producer's internal sequence
Stream-Seq: 10005                 # Application-level coordination (opt-in)

{"order": "..."}
```

Server validates both independently:
1. **Transport check**: Is `X-Producer-Seq: 42` valid for this producer/epoch? (dedup)
2. **Application check**: Is `Stream-Seq: 10005` > stream's `lastSeq`? (coordination)

---

### Producer Client

```typescript
interface IdempotentProducerOptions {
  stream: DurableStream

  // Pipelining configuration
  maxInFlight?: number        // Default: 5
  maxBatchBytes?: number      // Default: 1MB
  lingerMs?: number           // Default: 0 (send immediately when idle)

  // Callbacks
  onBatchAck?: (result: BatchResult) => void
  onError?: (error: Error, batch: PendingBatch) => void
}

class IdempotentProducer {
  constructor(options: IdempotentProducerOptions)

  // Fire-and-forget append (queued for batching)
  append(data: unknown, options?: AppendOptions): void

  // Wait for all pending batches to be acknowledged
  flush(): Promise<void>

  // Graceful shutdown
  close(): Promise<void>

  // Stats
  readonly pendingCount: number
  readonly inFlightCount: number
}
```

### Usage Example

```typescript
const producer = new IdempotentProducer({
  stream,
  maxInFlight: 5,
  onError: (error, batch) => {
    console.error(`Failed to write ${batch.count} messages:`, error)
    // Handle error (e.g., write to dead letter queue)
  }
})

// Fire-and-forget - producer handles batching and retries
for (const event of events) {
  producer.append(event)
}

// Wait for all writes to complete
await producer.flush()
```

### Server Protocol

**Request (client-provided producer ID, server-managed epoch):**
```http
POST /stream/my-stream HTTP/1.1
Content-Type: application/json
X-Producer-Id: order-service-1
X-Producer-Seq: 0

[{"event": "click"}, {"event": "scroll"}]
```

Note: Client sends `X-Producer-Id` and `X-Producer-Seq`. Server manages epochs internally.

**Response (success - first write or new epoch):**
```http
HTTP/1.1 201 Created
X-Stream-Offset: 00000042
X-Producer-Epoch: 1
```

**Response (success - subsequent write):**
```http
HTTP/1.1 201 Created
X-Stream-Offset: 00000043
```

**Response (duplicate - already received this seq):**
```http
HTTP/1.1 204 No Content
```

**Response (sequence gap - missing messages):**
```http
HTTP/1.1 409 Conflict
X-Expected-Seq: 5
X-Received-Seq: 7
```

**Response (stale epoch - zombie producer from old session):**
```http
HTTP/1.1 403 Forbidden
X-Error: stale-epoch
X-Current-Epoch: 2
```

---

### Server-Side State

```typescript
interface ProducerState {
  producerId: string
  epoch: number
  lastSeq: number
  lastUpdated: number  // For cleanup of stale producers
}

// Per-stream map of producer states
// Can be stored in-memory with periodic snapshots, or inline with messages
streamProducers: Map<string, ProducerState>
```

### Sequence Validation Logic

```typescript
function validateProducerWrite(
  stream: Stream,
  producerId: string,
  epoch: number,
  seq: number
): 'accept' | 'duplicate' | 'gap' | 'stale-epoch' {
  const state = stream.producers.get(producerId)

  if (!state) {
    // New producer, accept and initialize
    stream.producers.set(producerId, { producerId, epoch, lastSeq: seq })
    return 'accept'
  }

  if (epoch < state.epoch) {
    // Zombie producer from old epoch
    return 'stale-epoch'
  }

  if (epoch > state.epoch) {
    // New epoch, reset sequence tracking
    state.epoch = epoch
    state.lastSeq = seq
    return 'accept'
  }

  // Same epoch, check sequence
  if (seq <= state.lastSeq) {
    return 'duplicate'  // Already seen, idempotent ack
  }

  if (seq > state.lastSeq + 1) {
    return 'gap'  // Missing messages, reject
  }

  // seq === state.lastSeq + 1
  state.lastSeq = seq
  return 'accept'
}
```

### Producer ID Assignment

**Option A: Server-assigned (like Kafka)**
```http
POST /stream/my-stream/producer HTTP/1.1

Response:
{
  "producerId": "p-abc123",
  "epoch": 1
}
```

Downside: Requires 1 RTT before first write. For low-latency use cases, this overhead is significant.

**Option B: Client-provided with server-managed epochs (like Pulsar)** ✅ RECOMMENDED
```http
POST /stream/my-stream HTTP/1.1
X-Producer-Id: order-service-1
X-Producer-Seq: 0

Response:
HTTP/1.1 201 Created
X-Producer-Epoch: 1
```

**Why client-provided is better for durable-streams:**

1. **Zero RTT to first byte**: Producer can send data immediately, no handshake required
2. **Simpler protocol**: No separate producer registration endpoint
3. **Better debugging**: Meaningful names like `order-service-1` vs opaque IDs like `p-abc123`
4. **Stateless clients**: Client doesn't need to persist server-assigned ID

**How epochs work with client-provided IDs:**

- Server tracks `(producerId, epoch, lastSeq)` per stream
- On first write from a producer ID: epoch=1, accept
- On write with seq=0 (restart): bump epoch, reset sequence tracking
- Old epoch writes rejected as stale (zombie fencing)

```typescript
// Client generates stable ID (e.g., from config, hostname, or UUID persisted to disk)
const producerId = process.env.PRODUCER_ID ?? `${hostname()}-${processId}`

// First write includes producer ID, server assigns epoch
producer.append(data)  // X-Producer-Id: web-server-1, X-Producer-Seq: 0
                       // Response: X-Producer-Epoch: 1

// Subsequent writes in same session
producer.append(data)  // X-Producer-Id: web-server-1, X-Producer-Seq: 1
```

**Epoch bump on restart:**
```
Session 1: seq 0, 1, 2, 3... (epoch 1)
[crash/restart]
Session 2: seq 0 → server sees seq reset, bumps to epoch 2
           seq 1, 2, 3... (epoch 2)

If zombie from session 1 sends seq=4 with epoch=1 → rejected (stale epoch)
```

### Pipelining Implementation

```typescript
class IdempotentProducer {
  #inFlight: Map<number, PendingBatch> = new Map()
  #nextSeq: number = 0
  #maxInFlight: number = 5

  async #sendBatch(batch: QueuedMessage[]): Promise<void> {
    // Wait if at max in-flight
    while (this.#inFlight.size >= this.#maxInFlight) {
      await this.#waitForAck()
    }

    const seq = this.#nextSeq++
    const pending = { seq, batch, promise: new Deferred() }
    this.#inFlight.set(seq, pending)

    // Send without waiting (pipelining)
    this.#doSend(pending).catch(err => this.#handleError(err, pending))
  }

  async #doSend(pending: PendingBatch): Promise<void> {
    const response = await fetch(this.#url, {
      method: 'POST',
      headers: {
        'X-Producer-Id': this.#producerId,
        'X-Producer-Seq': String(pending.seq),
      },
      body: JSON.stringify(pending.batch.map(m => m.data))
    })

    if (response.status === 201 || response.status === 204) {
      // Success or duplicate (both are fine)
      this.#inFlight.delete(pending.seq)
      pending.promise.resolve()
    } else if (response.status === 409) {
      // Gap - missing messages, surface error to application
      this.#handleGap(pending, response)
    } else if (response.status === 403) {
      // Stale epoch - this producer session is zombied
      // Another instance with same ID started a new session
      this.#handleStaleEpoch(pending, response)
    }
  }
}
```

---

### Phase 1: Add Idempotent Producer (Non-Breaking)

1. Add `IdempotentProducer` class alongside existing `DurableStream`
2. Add server-side producer state tracking
3. Existing code continues to work unchanged

### Phase 2: Deprecate Direct Batching

1. Mark `DurableStream.append()` with batching as deprecated
2. Encourage migration to `IdempotentProducer`
3. Update documentation

### Phase 3: Default Behavior

1. Consider making idempotent producer the default
2. Similar to Kafka 3.0 making `enable.idempotence=true` the default

---

## Open Questions

1. **Producer state storage**: In-memory with snapshots vs. inline with messages?
   - Kafka stores inline (sequence in message metadata)
   - Trade-off: Memory vs. recovery complexity

2. **Producer ID lifetime**: How long to retain producer state?
   - Kafka: Until log segment is deleted
   - Could use TTL with periodic refresh

3. ~~**Epoch assignment**: Server-assigned vs. client-provided?~~ **RESOLVED**
   - **Decision**: Client-provided IDs with server-managed epochs (Pulsar-style)
   - **Rationale**: Zero RTT to first byte is critical for low-latency use cases

4. **Gap handling**: Reject vs. buffer-and-reorder?
   - Kafka rejects (client retries)
   - Simpler server, pushes complexity to client

---

### Industry Documentation

- [KIP-98: Exactly Once Delivery and Transactional Messaging](https://cwiki.apache.org/confluence/display/KAFKA/KIP-98+-+Exactly+Once+Delivery+and+Transactional+Messaging)
- [Confluent: Exactly-once Semantics in Kafka](https://www.confluent.io/blog/exactly-once-semantics-are-possible-heres-how-apache-kafka-does-it/)
- [Pulsar PIP-6: Guaranteed Message Deduplication](https://github.com/apache/pulsar/wiki/PIP-6:-Guaranteed-Message-Deduplication)
- [RabbitMQ Streams: Message Deduplication](https://www.rabbitmq.com/blog/2021/07/28/rabbitmq-streams-message-deduplication)
- [NATS JetStream Model Deep Dive](https://docs.nats.io/using-nats/developer/develop_jetstream/model_deep_dive)
- [Redis Streams Documentation](https://redis.io/docs/latest/develop/data-types/streams/)
- [Google Pub/Sub Exactly-Once Delivery](https://cloud.google.com/pubsub/docs/exactly-once-delivery)
- [Azure Event Hubs Idempotent Producer Design](https://gist.github.com/jsquire/caeacd165785a35811e00292203eb36d)

### Academic Papers

- [MillWheel: Fault-Tolerant Stream Processing at Internet Scale](https://research.google/pubs/millwheel-fault-tolerant-stream-processing-at-internet-scale/) (Google, 2013)
- [Lightweight Asynchronous Snapshots for Distributed Dataflows](https://arxiv.org/abs/1506.08603) (Flink checkpointing)
- [Impossibility of Distributed Consensus with One Faulty Process](https://groups.csail.mit.edu/tds/papers/Lynch/jacm85.pdf) (FLP, 1985)

### Blog Posts & Analysis

- [Jay Kreps: Exactly-once Support in Apache Kafka](https://medium.com/@jaykreps/exactly-once-support-in-apache-kafka-55e1fdd0a35f)
- [The Impossibility of Exactly-Once Delivery](https://blog.bulloak.io/post/20200917-the-impossibility-of-exactly-once/)
- [Flink Checkpointing Internals](https://nightlies.apache.org/flink/flink-docs-release-1.8/internals/stream_checkpointing.html)

### Related Issues & PRs

- [PR #51: Kafka-style Idempotent Producers](https://github.com/durable-streams/durable-streams/pull/51)
- [Issue #32: OCC Support](https://github.com/durable-streams/durable-streams/issues/32)
- [Issue #5: WebSocket API](https://github.com/durable-streams/durable-streams/issues/5)

KyleAMathews:
## Executive Summary

Implements Kafka-style idempotent producers with exactly-once write semantics across all clients (TypeScript, Go, Python). Users get fire-and-forget writes that batch automatically, pipeline up to 5 concurrent requests, and deduplicate retries server-side—turning "at-least-once" into "exactly-once" without application complexity.

Fixes #108

---

### Approach

**Server-side deduplication via (producerId, epoch, seq) headers:**

```
POST /v1/stream/events
Producer-Id: order-service-1
Producer-Epoch: 0
Producer-Seq: 42
Content-Type: application/json

[{"orderId": "abc", "status": "shipped"}]
```

The server tracks `(producerId → {epoch, lastSeq})` per stream. Duplicate sequences return `204 No Content` (idempotent success). Stale epochs get `403 Forbidden` (zombie fencing). Sequence gaps return `409 Conflict` with expected/received headers for client retry.

**Client pipelining with automatic batching:**

```typescript
// Fire-and-forget - batched and sent in background
for (const event of events) {
  producer.append(event)  // Don't await
}
await producer.flush()  // Ensure delivery before shutdown
```

Messages accumulate for `lingerMs` (default 5ms) or until `maxBatchBytes` (1MB), then ship as a single HTTP request. Up to `maxInFlight` (5) batches fly concurrently.

### Key Invariants

1. **Sequence monotonicity**: Server rejects seq gaps (409) and stale epochs (403)
2. **Exactly-once per batch**: Entire HTTP body is atomic—either all values stored or none
3. **Epoch fencing**: Once epoch N+1 appears, epoch N writes are permanently rejected
4. **At-most-one writer per producerId**: autoClaim ensures single active producer

### Non-goals

- **Transactions across streams**: Each stream tracks producer state independently
- **Server-assigned producer IDs**: Client provides IDs (zero RTT overhead)
- **Guaranteed ordering across producers**: Only within single producer's sequence

### Trade-offs

| Decision | Alternative | Rationale |
|----------|-------------|-----------|
| Client-declared epochs | Server-assigned | Zero handshake RTT, simpler protocol |
| Per-batch sequences | Per-message | Matches HTTP request atomicity, fewer headers |
| In-memory producer state | Persisted | Simplicity; epoch bump handles restarts |
| 5 max in-flight | Unbounded | Balance throughput vs memory; user-configurable |

---

## Verification

```bash
# Run all tests
pnpm test:run

# Run specific client conformance tests
pnpm test:run -- --client python
pnpm test:run -- --client go

# Run server conformance tests
pnpm test --filter @durable-streams/server-conformance-tests

# Run Go tests directly
cd packages/caddy-plugin && go test ./...
```

**Benchmark results** (10k appends × 100 bytes):

| Client | Before | After | Improvement |
|--------|--------|-------|-------------|
| Go | 656k ops/s | 5.6M ops/s | **8.6x** |
| Python | 187k ops/s | 501k ops/s | **2.7x** |
| TypeScript | ~700k ops/s | ~760k ops/s | ~9% |

### Files Changed

**Protocol & Server:**
- `PROTOCOL.md` - New headers, status codes, producer state semantics
- `packages/server/src/store.ts` - Producer state tracking (epoch, lastSeq per producerId)
- `packages/server/src/server.ts` - Header parsing, duplicate detection, error responses
- `packages/caddy-plugin/handler.go` - Same logic for Go server
- `packages/caddy-plugin/store/*.go` - Producer state in memory/file stores

**Clients:**
- `packages/client/src/idempotent-producer.ts` - TypeScript: batching, pipelining, autoClaim
- `packages/client-go/idempotent_producer.go` - Go: same features, goroutine-based
- `packages/client-py/src/durable_streams/idempotent_producer.py` - Python: asyncio-based

**Tests:**
- `packages/server-conformance-tests/src/index.ts` - Protocol compliance (409 gaps, 403 stale, 204 dupe)
- `packages/client-conformance-tests/test-cases/producer/idempotent/` - Kafka-inspired test structure

**Examples:**
- `examples/state/wikipedia-worker/` - Updated to IdempotentProducer
- `examples/test-ui/` - Updated to IdempotentProducer

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -626,7 +626,7 @@ const stream = await DurableStream.create({
 
 // Use IdempotentProducer for reliable, exactly-once token delivery
 let fenced = false
-const producer = new IdempotentProducer(stream, `llm-worker-${workerId}`, {
+const producer = new IdempotentProducer(stream, generationId, {
   autoClaim: true, // Auto-recover if another worker took over
   lingerMs: 10, // Send batches every 10ms for low latency
   onError: (error) => {


samwillis (changes requested):
I had both Open4.5 and GPT5.2 review this PR as individual subsystems and the protocol change, then check each others review, before finally having a combined review constructed - see below

I think once the things below are done this is good to go. I am happy with the protocol design myself.

-----

# PR #140: Kafka-style Idempotent Producers — Subsystem-by-Subsystem Review

**Scope:** This review covers the protocol changes, each server implementation, each client implementation (TypeScript/Go/Python), and the conformance tests/harness introduced or modified by PR #140 (“Implement Kafka-style idempotent producers”).

**Review basis:** This document is written against the **current committed state of branch** `claude/implement-idempotent-producers-oBro6` (working tree clean at time of writing). Where relevant, I also note **recommended changes** that are *not currently implemented*.

---

## Executive summary

This PR adds Kafka-style idempotent producers using `(Producer-Id, Producer-Epoch, Producer-Seq)` headers to provide **exactly-once semantics per producer-batch** under retries and network errors *as long as the server persists producer state correctly and clients follow monotonic seq rules per epoch*.

The design is strong and the broad implementation coverage (TS server + Go server + three clients + conformance tests) is impressive. However, there are two high-impact correctness issues in the current branch state:

1. **CRITICAL (clients): `autoClaim` + `maxInFlight>1` sequence reset race**  
   All three clients reset their local `nextSeq` on 403 auto-claim. If multiple batches were already assigned sequence numbers before the 403 resolves, the reset can cause subsequent batches to reuse already-consumed sequence numbers, producing **persistent 409 conflicts** and “stuck producer” behavior.

2. **HIGH (Caddy server stores): `Stream-Seq` validation happens before producer dedupe**  
   In both `packages/caddy-plugin/store/memory_store.go` and `file_store.go`, the store checks `Stream-Seq` before it checks producer dedupe. That can turn a safe retry (which should return **204 duplicate**) into a **409 Stream-Seq conflict**, breaking the intended “transport idempotency outranks application sequencing” layering.

There are also several “spec clarity” and “harness coverage” issues that are not merge-blocking individually but matter for correctness long-term.

---

## Severity legend used in this review

- **CRITICAL**: can cause lost availability / stuck writes / incorrect semantics in plausible production usage.
- **HIGH**: violates protocol/client-server contract or creates cross-implementation divergence.
- **MEDIUM**: correctness edge-case, or significant operability issue under load.
- **LOW**: cleanup, ergonomics, doc clarity.

---

## 1) Protocol specification (`PROTOCOL.md`)

### What the protocol adds
Section **5.2.1 Idempotent Producers** defines:
- Request headers: `Producer-Id`, `Producer-Epoch`, `Producer-Seq` (all-or-none)
- Response headers: `Producer-Epoch`, `Producer-Seq`, `Producer-Expected-Seq`, `Producer-Received-Seq`
- Status codes: 200 (accepted), 204 (duplicate), 400 (bad headers), 403 (stale epoch), 409 (gap)
- Validation pseudocode, restart flow, auto-claim flow
- Requirements: **serialize per (stream,producerId)**; **atomicity** of append + producer state for persistent stores

### Strengths
- **Two-layer sequencing design** (transport vs application) is correct and future-proof:
  - Transport idempotency: `(producerId, epoch, seq)` for retry safety
  - Application ordering: `Stream-Seq` for cross-restart ordering/coordination
- **Per-batch seq** matches HTTP request atomicity (one request commits all-or-none).
- **403 fencing** is correct: stale epochs must be rejected permanently once a higher epoch is observed.
- **409 gap** with expected/received headers is essential for pipelined clients when requests arrive out-of-order.

### Spec gaps / ambiguities
- **Partial producer headers** (**MEDIUM**): The spec says “all three MUST be provided together” but does not explicitly state:
  - The required response status should be **400 Bad Request** for partial headers.
  - Current implementations and conformance tests assume 400.
- **Empty `Producer-Id`** (**MEDIUM**): Not explicitly forbidden in the spec. Implementations reject empty `Producer-Id` with 400 and tests assert that behavior. The spec should say “MUST be a non-empty string”.
- **Numeric bounds** (**MEDIUM**): Epoch/seq are “integers” but the spec doesn’t define a max range. This matters because:
  - TypeScript implementations must guard `Number.MAX_SAFE_INTEGER`.
  - Go/Python can handle larger values, but interoperable semantics require a shared bound or a string/bigint encoding rule.
- **Atomicity requirement vs known implementations** (**MEDIUM**): The protocol says persistent stores “MUST commit producer state updates and log appends atomically”. Some stores document non-crash-atomic behavior today. The spec should either:
  - soften to SHOULD + require documenting failure modes, or
  - keep MUST and treat non-atomic stores as non-compliant, with explicit recovery guidance (epoch bump).

### Recommendation
Make the spec explicit about:
- Partial header behavior: **400**
- `Producer-Id` must be **non-empty**
- Epoch/seq max (likely `<= 2^53-1` unless adopting bigint strings)
- If non-atomic persistent stores are allowed, specify the acceptable failure mode and client recovery behavior.

---

## 2) TypeScript reference server (`packages/server`)

### Goals of this subsystem
This is the reference server used by tests and conformance suites. It sets the “normative behavior” for protocol interpretation.

### Implementation highlights
- **Strict producer header parsing and validation** in `src/server.ts`
  - digit-only parsing
  - `Number.isSafeInteger()` to reject unsafe integers
  - all-or-none header enforcement
  - empty producer id rejected
- **Producer validation is non-mutating and committed only after append succeeds** in `src/store.ts`
  - Prevents “producer state advanced even though append failed” (critical for JSON validation errors).
- **Per-(stream,producerId) serialization** exists via an in-memory lock.

### Correctness notes
- The ordering of checks in the TS store is important and correct:
  - Producer validation/dedupe happens before `Stream-Seq` checks so retries are idempotent at the transport layer even if an application seq would now conflict.
- State TTL cleanup for producers is implemented (in-memory, 7 days). This is acceptable for test/reference use.

### Potential issues / improvements
- **Producer cleanup cost** (**LOW/MEDIUM**): cleanup iterates producer map on access. If a single stream accumulates very large producer cardinality, this becomes O(n) per append. A periodic or probabilistic cleanup strategy would scale better.
- **Persistent-store atomicity** (**MEDIUM**): the TS file-backed store reconciles offsets on recovery but producer state is not log-derived, so crash windows can violate “exactly once” unless the store is transactionally atomic.

### Verdict
**Approve** (reference-quality). The core algorithmic design is correct and is a good model for other servers.

---

## 3) Go/Caddy production server (`packages/caddy-plugin`)

### 3.1 Handler layer (`handler.go`)
Strengths:
- Correct parsing/validation of producer headers and mapping to status codes.
- SSE CR/LF injection hardening is present.
- Security headers are set as per protocol’s browser guidance.

Risks:
- Mostly in the store layer, not in handler logic.

### 3.2 Store layer (`store/memory_store.go`, `store/file_store.go`)

#### CRITICAL ordering issue: Stream-Seq checked before producer dedupe (**HIGH**)
In both memory and file store `Append()` paths, the code currently does:
1) Validate `Stream-Seq` regression first  
2) Then validate producer headers / dedupe

Example (file store):
```go
// Validate sequence number (Stream-Seq - application layer)
if opts.Seq != "" {
  if meta.LastSeq != "" && opts.Seq <= meta.LastSeq {
    return AppendResult{}, ErrSequenceConflict
  }
}

// Validate producer (if headers provided)
...
if result.ProducerResult == ProducerResultDuplicate {
  return AppendResult{ ProducerResult: ProducerResultDuplicate }, nil
}
```

**Why this matters:** A retry that should be safely deduped as **204** can be rejected as a **Stream-Seq conflict** if the caller also set `Stream-Seq` (common when applications use Stream-Seq for cross-process ordering). The TS server explicitly avoids this by checking producer dedupe first.

**Recommendation:** Move `Stream-Seq` validation *after* producer validation, and ensure the duplicate path returns before any Stream-Seq conflict logic.

#### Crash-atomicity for file store (**MEDIUM/HIGH**)
Both reviewers agree this is a known limitation:
- Data is appended to a file, producer state is stored separately in bbolt.
- A crash between these steps can leave producer state stale; a retry might be re-accepted and duplicate data appended.

This is documented as a known limitation (good), but be careful with wording such as “file is source of truth” because the file does not encode producer state.

#### Concurrency
Per-producer mutex serialization exists. That is required by the protocol.

### Verdict
**Request changes** until the `Stream-Seq` vs producer-dedupe ordering issue is fixed (HIGH impact interop/correctness).

---

## 4) TypeScript client (`packages/client`)

### What was added
`IdempotentProducer` implements:
- fire-and-forget append() with batching
- pipelining with `maxInFlight`
- handling 409 gaps by waiting for earlier sequences then retrying
- optional auto-claim on 403

### Strengths
- **409 reordering coordination** is correctly designed: when server returns expectedSeq < our seq, wait for earlier seq completions then retry that same seq. This is the right model for out-of-order arrival when `maxInFlight>1`.
- **JSON batching semantics** are correct: send arrays in JSON mode so the server flattens exactly one level.
- **Separation of concerns**: producer headers are handled in producer; stream append remains simpler.

### CRITICAL: autoClaim sequence reset race (**CRITICAL**)
Current TS client behavior on 403 with auto-claim:
```ts
this.#epoch = newEpoch
this.#nextSeq = 1 // This batch will use seq 0
return this.#doSendBatch(batch, 0, newEpoch)
```

**Why it is a real bug:** It can create seq reuse after a successful auto-claim when multiple batches were already allocated sequence numbers before the 403 resolves (especially with `maxInFlight>1` and appends happening quickly). This yields persistent 409 “expected higher seq, received old seq” and effectively wedges the producer.

**Why current tests don’t catch it:** existing autoclaim tests either:
- don’t generate 403 at all, or
- use a single append, or
- set `maxInFlight=1`.

**Recommendation options (in increasing complexity):**
- **Option 1 (short-term safety):** Disallow `autoClaim` with `maxInFlight>1` (throw or force `maxInFlight=1`), and document it clearly.
- **Option 2 (correct behavior):** Ensure that when auto-claim occurs, all already-allocated seqs are consistent with the new epoch and `nextSeq` is advanced to the maximum allocated + 1. That requires tracking “highest allocated seq” independent of `nextSeq` resets, or rewriting queued tasks.

### Additional notes
- The inline comment “This batch will use seq 0” is misleading; the batch already has a seq, and the code is retrying as seq 0. The comment should describe the intended invariant instead of restating the immediate action.
- No general retry/backoff for network errors/5xx: acceptable if documented, but don’t oversell “network resilience” beyond protocol-level dedupe.

### Verdict
**Request changes** until auto-claim race is fixed or the feature combination is explicitly prevented.

---

## 5) Go client (`packages/client-go`)

### Strengths
- Mirrors TS design: batching, pipelining, 409 coordination, optional auto-claim.
- Error types are explicit and unwrap to sentinel errors.

### CRITICAL: same autoClaim seq reset race (**CRITICAL**)
Same pattern exists:
```go
p.epoch = newEpoch
p.nextSeq = 1 // This batch uses seq 0
return p.doSendBatch(ctx, batch, 0, newEpoch)
```

This is the same underlying bug: with queued batches already assigned seq>0, resetting nextSeq can lead to seq reuse later.

### Other notes
- Fire-and-forget API limits cancellation control (no context on Append); acceptable but worth documenting.

### Verdict
**Request changes** until auto-claim race is fixed or disallowed with concurrency.

---

## 6) Python client (`packages/client-py`)

### Strengths
- Asyncio design is appropriate and consistent with TS/Go semantics.
- Uses httpx AsyncClient; supports ownership patterns and context manager usage.

### CRITICAL: same autoClaim seq reset race (**CRITICAL**)
Same pattern exists:
```py
self._epoch = new_epoch
self._next_seq = 1  # This batch will use seq 0
return await self._do_send_batch(batch, 0, new_epoch)
```

### Python-specific improvements
- Use `asyncio.get_running_loop()` instead of deprecated `get_event_loop()` in modern Python (LOW/MEDIUM).
- Flush logic is non-trivial; consider adding a focused unit test for “flush while callbacks schedule additional batches”.

### Verdict
**Request changes** until auto-claim race is fixed or disallowed with concurrency.

---

## 7) Conformance tests & harness

### 7.1 Client conformance (`packages/client-conformance-tests`)
Strengths:
- The YAML suite coverage for idempotent producers is excellent: epoch fencing, gaps, dedupe semantics, concurrency, JSON batching, multi-producer isolation.

Issues:
- **A YAML assertion is currently not enforced** (**MEDIUM**): at least one test uses `expect.dataExact`, but the runner only validates `data`, `dataContains`, `dataContainsAll`, etc. As a result, that specific ordering assertion is silently ignored and the test may pass even if the behavior is wrong.

Missing tests (not exhaustive):
- Auto-claim 403 while multiple batches already allocated seqs with `maxInFlight>1` (**should be added; this catches the CRITICAL client bug**).
- Numeric boundary tests around MAX_SAFE_INTEGER for TS implementations.
- Producer state TTL expiration semantics (if any store cleans up producer state).

### 7.2 Server conformance (`packages/server-conformance-tests`)
Strengths:
- Very thorough suite including property-based tests and idempotent producer cases.

Practical issue in monorepo execution (**MEDIUM**):
- The package has its own CLI, but in a monorepo with a root vitest multi-project config, invoking the CLI may fail to discover tests unless special care is taken. (This is an execution ergonomics issue rather than protocol correctness.)

---

## Cross-cutting: “exactly once” vs persistent store reality

The protocol promises exactly-once semantics, but persistent implementations must meet a high bar:
- If producer state and append are not committed atomically, a crash can reintroduce duplicates on retry.
- Some implementations document that epoch bump can recover, but that changes the semantics from “exactly once even across crash windows” to “exactly once as long as you don’t crash between the data write and state update”.

**Recommendation:** Decide whether the ecosystem treats this as:
- strict MUST compliance (and non-atomic stores are non-compliant), or
- allowed-but-documented tradeoff with explicit client recovery guidance.

---

## Required changes before merge (recommended)

1. **Fix client auto-claim race** (TS + Go + Py)  
   Either implement correct seq accounting across queued batches, or disallow `autoClaim` with `maxInFlight>1` until the full fix lands.

2. **Fix Caddy store ordering** (file + memory)  
   Validate producer/dedupe first, return duplicates early, then apply `Stream-Seq` checks only for accepted non-duplicate appends.

3. **Harden conformance harness**  
   Add `dataExact` validation to the runner and add the missing auto-claim+concurrency+403 test.

---

## Appendix: Notable code locations

- Protocol: `PROTOCOL.md` Section 5.2.1
- TS server producer validation/locking: `packages/server/src/store.ts`
- TS server producer header parsing: `packages/server/src/server.ts`
- Caddy store ordering issue:
  - `packages/caddy-plugin/store/file_store.go`
  - `packages/caddy-plugin/store/memory_store.go`
- Client auto-claim seq reset:
  - TS: `packages/client/src/idempotent-producer.ts`
  - Go: `packages/client-go/idempotent_producer.go`
  - Py: `packages/client-py/src/durable_streams/idempotent_producer.py`

samwillis (approved):
All good :shipit: 

-----

# Re-Review Confirmation: PR #140 After Rebase

## ✅ All Critical Fixes Remain Intact

After verifying the codebase post-rebase, I confirm that:

### 1. **autoClaim + maxInFlight Race Fix** — Still Present ✅

All three clients have the `epochClaimed` protection mechanism:

| Client | Flag | Check Location |
|--------|------|---------------|
| **TypeScript** | `#epochClaimed` | Line 386: blocks pipelining until claimed |
| **Go** | `epochClaimed` | Line 493: blocks pipelining until claimed |
| **Python** | `_epoch_claimed` | Line 419: blocks pipelining until claimed |

### 2. **Caddy Store Ordering Fix** — Still Present ✅

Both stores validate producer headers FIRST, before Stream-Seq:
- `memory_store.go` line 275: "Validate producer FIRST"
- `file_store.go` line 390: "Validate producer FIRST"

### 3. **TS Server Ordering** — Correct ✅
- `store.ts` line 411: "Handle producer validation FIRST"
- `file-store.ts` line 708: "Handle producer validation FIRST"

## ✅ All Tests Pass

| Client | Tests | Status |
|--------|-------|--------|
| TypeScript | 718 passed, 3 skipped | ✅ |
| Go | 718 passed, 3 skipped | ✅ |
| Python | 718 passed, 3 skipped | ✅ |
| Caddy Go tests | All pass | ✅ |

## Summary

**The rebase did not break anything.** All critical fixes from the previous review remain intact, and all conformance tests pass across all three client implementations.

**Verdict: APPROVE ✅**

The PR is ready for merge.

PR #144, opened 2026-01-06
cli: fix JSON mode to properly parse input.

thruflo:
## Summary

Adds JSON modes to the CLI `write` command with `--json` and `--batch-json` flags. Users can now write JSON data to streams as a single message or as multiple messages from an array.

---

## Approach

**New CLI flags for write command:**
- `--json` - Write JSON input as a single message
- `--batch-json` - Write JSON array input as multiple messages (each element stored separately)

**Flag behavior comparison:**

| Flag | Input `[{}, {}]` | Result |
|------|------------------|--------|
| `--json` | `[{}, {}]` | 1 message: `[{}, {}]` |
| `--batch-json` | `[{}, {}]` | 2 messages: `{}`, `{}` |

Use `--json` when you want the exact JSON value stored. Use `--batch-json` when your input is an array of messages to append individually.

**Implementation:**
- `parseWriteArgs.ts` - Argument parser with unknown flag rejection
- `jsonUtils.ts` - Content-type detection and generator-based array flattening
- Main CLI updated to handle JSON parsing and batched appends

**Key invariants:**
1. `--batch-json` implies JSON content-type (no need to specify `--json` alongside it)
2. Unknown flags throw errors (fail-fast vs silent ignore)
3. Flags can appear before or after content

**Non-goals:**
- No support for `application/json-patch+json` or other JSON variants in `isJsonContentType`
- No recursive array flattening (one level only, by design)

---

## Other changes

- **Callback signature updates**: Converted `async (chunk) => {...}` to `(chunk) => {...; return Promise.resolve()}` across subscriber callbacks (benchmarks, stream-store, stream-db, conformance tests)
- **README**: Added Java community implementation link

---

## Verification

```bash
# Run CLI tests (33 tests)
pnpm --filter @durable-streams/cli test

# Manual testing
durable-stream write my-stream '{"foo": 1}' --json
durable-stream write my-stream '[{"a":1},{"b":2}]' --batch-json
```

---

## Files changed

| File | Change |
|------|--------|
| `packages/cli/src/index.ts` | Main CLI with JSON mode handling |
| `packages/cli/src/parseWriteArgs.ts` | Argument parser |
| `packages/cli/src/jsonUtils.ts` | JSON utilities |
| `packages/cli/test/*.test.ts` | Unit tests (33 total) |
| `packages/cli/README.md` | Documentation |
| `README.md` | Java community link + typo fix |
| `examples/test-ui/*` | Callback signature updates |
| `packages/benchmarks/*` | Callback signature updates |
| `packages/state/*` | Callback signature updates |

### DIFF
--- a/.changeset/cli-json-flags.md
+++ b/.changeset/cli-json-flags.md
@@ -7,4 +7,4 @@ Add JSON support to CLI write command with two modes:
 - `--json`: Write JSON input as a single message
 - `--batch-json`: Write JSON array input as multiple messages (each element stored separately)
 
-Also adds `--content-type <type>` for arbitrary content types, and improved error handling to reject unknown flags.
+Also improved error handling to reject unknown flags.


KyleAMathews:
It's unexpected that an array would get unwrapped in this context. That behavior is just so we can batch up JSON writes when sending them to the server. If piping, I'd expect the actual data I send to be preserved as is. If people want actual want to pipe a batch of data, perhaps that should be opt-in? E.g. a different flag `--batch-json`?

thruflo:
I wasn’t sure but I landed on this as it felt equivalent to the way other usage works. As in, you post an array to the server it gets one-level unwrapped. Here, you cat to the CLI it gets one level unwrapped. If you want to write an array, you wrap it with another array. As you would posting to the server.

I can totally see not doing that. Or adding a —batch flag. Let me know what you’d prefer.

KyleAMathews (approved):

PR #150, opened 2026-01-08
Use offset=now for presence stream in y-durable-streams

balegas:
## Summary
- Modified the y-durable-streams provider to use `offset=now` for the awareness (presence) stream
- Broadcasts our presence immediately before starting to listen, ensuring other clients see us right away
- Uses `offset=now` with long-poll to skip historical presence data and wait only for new updates
- Historical presence data is not relevant - we only care about who is currently online

## How it works
1. Client connects to presence stream
2. **Immediately** broadcasts its own presence and starts heartbeat
3. Starts long-poll with `offset=now` to wait for other clients' presence updates
4. Receives other clients' presence when they heartbeat (every 15s)

This approach:
- Saves bandwidth by not downloading potentially large presence history
- Ensures our presence is visible to others immediately
- Uses a single request instead of catch-up + long-poll

## Test plan
- [x] Typecheck passes
- [x] Manual testing with yjs-demo - presence works correctly with multiple clients

### DIFF
--- a/.changeset/fuzzy-buckets-nail.md
+++ b/.changeset/fuzzy-buckets-nail.md
@@ -1,0 +1,5 @@
+---
+"@durable-streams/y-durable-streams": patch
+---
+
+Use offset=now in presence stream


packages/y-durable-streams/src/y-durable-streams.ts
```diff
@@ -369,10 +369,18 @@ export class DurableStreamsProvider extends ObservableV2<DurableStreamsProviderE
...
 
-    // Start streaming from the beginning
+    // Broadcast our presence immediately before starting to listen.
+    // This ensures other clients see us right away, even before we receive their updates.
+    this.awarenessReady = true
+    this.broadcastAwareness()
+    this.startAwarenessHeartbeat()
+
+    // Start streaming from current position - we don't need historical presence data.
+    // Use live: "auto" so the first request (catch-up) returns immediately,
```
KyleAMathews: this isn't true — long-poll returns immediately too — LLM got confused by the name

packages/y-durable-streams/src/types.ts
```diff
@@ -69,6 +69,14 @@ export interface DurableStreamsProviderOptions {
...
    */
   debug?: boolean
+
+  /**
+   * The live mode to use for streaming updates.
+   * - "long-poll": Use long-polling for live updates
+   * - "sse": Use Server-Sent Events for live updates
+   * @default "long-poll"
+   */
+  liveMode?: Extract<LiveMode, `long-poll` | `sse`>
```
KyleAMathews: it has to be `long-poll` btw since yjs is a binary protocol. SSE is only for text or json

balegas: ah never thought of that. reverting commit

PR #183, opened 2026-01-20
Add "/" to list of banned characters

broady:
This character was already banned by the conformance tests, the spec just diverged.

Additionally, I changed it to list out the actual bytes in code tags - let me know if you want me to write out "forward slash" instead. Or maybe we should list out ASCII codes? I think what I've got here is fine.


KyleAMathews (approved):
Yeah listing the actual characters is more legible for sure. Thanks for the fix!

PR #220, opened 2026-01-27
feat: implement stream closure for servers and clients

### LINKED ISSUE #215: Add explicit stream closure / EOF semantics
### Summary

Durable Streams currently model an append-only log that can be read indefinitely. For some use cases we need **finite streams**: an explicit “end of stream” (EOF) state that is **durably recorded**, and that **readers can observe** so they know there is nothing more to read.

### Why this matters

Without an explicit close/EOF:

* Readers can’t distinguish **“no data yet”** from **“no more data ever”**.
* Clients are forced into heuristics (timeouts, polling, extra side-channels) to decide a stream is finished.
* A common workaround today is for writers to append a **sentinel “end message”** into the stream, have readers watch for it, and then close their own read connections. This works, but it pushes protocol semantics into application payloads, adds branching and edge cases (what if the sentinel is missed, duplicated, or not applicable for binary streams?), and becomes repetitive boilerplate for a very common “finite stream” pattern.

For systems that depend on completion semantics (e.g. “this request finished”, “this job output is done”), this is a correctness and UX issue, not just a convenience feature.

### Primary motivating use case: fetch proxy

We’re building a “fetch proxy” on top of Durable Streams:

1. Client supplies a URL (and request params).
2. Server fetches the upstream URL and **streams the upstream response body into a Durable Stream** as it arrives.
3. When the upstream request ends, the server must **close the Durable Stream**.
4. Any clients reading the stream should be able to observe EOF and conclude:

   * the upstream request has ended, and
   * there will be no more bytes/events to read.

In this case, “stream closed” is equivalent to “upstream response completed”, and readers need that signal to stop waiting and finalize (e.g. finish rendering, compute hashes, persist artifacts, trigger next steps).

### Requirements / expected semantics

**Writer side**

* Provide an explicit operation to transition a stream into **CLOSED** (EOF) state.
* Closing must be **durable** and **monotonic** (once closed, it stays closed).
* Closing should be **idempotent** (closing an already-closed stream should succeed or return a stable “already closed” response).
* After closure, **no further appends are allowed** (server rejects writes with a clear error).
* It would be useful to support an **atomic “append final message and close”** operation, so writers can emit a final chunk/event (e.g. summary, status, footer, trailers) and guarantee that:

  * the final append is the last item in the stream, and
  * the stream transitions to CLOSED in the same step (no race where another writer appends after the “final” message, and no gap where readers might see the final message but the stream is still open).

**Reader side**

* Readers must be able to detect closure and treat it as EOF.
* If a reader has consumed up to the current end and the stream is closed, the server should respond in a way that allows the client to stop reading immediately (no long-hanging connections, no ambiguity).
* Ideally, readers can learn:

  * that the stream is closed, and
  * the final offset/position at which it was closed (so they can confirm they have all data).

### Backwards compatibility

* Existing open-ended streams should continue to work unchanged.
* Writers that never close streams should not be affected.

### What success looks like

* Fetch-proxy streams can be written and then closed so that:

  * all readers can deterministically observe completion,
  * the server rejects further writes after close,
  * clients don’t need timeouts or side channels to decide the stream is finished.

samwillis:
**Stacked on #217, but base set to main so CI runs**

Also incorporate:
- #221 
- #222

----

## Stream Closure (EOF) Support

Implements stream closure functionality as specified in PROTOCOL.md Section 4.1, allowing writers to explicitly signal "no more data ever" to readers.

### Scope

This PR implements stream closure for:
- **TypeScript reference server** (`packages/server`)
- **TypeScript client library** (`packages/client`)
- **Server conformance tests** (`packages/server-conformance-tests`)
- **Client conformance tests** (`packages/client-conformance-tests`)

Other clients (Python, Go) and servers (Caddy plugin) are not included in this PR.

### Protocol

The `Stream-Closed: true` header signals stream closure:
- **Create closed**: `PUT` with `Stream-Closed: true` creates a stream in closed state
- **Close stream**: `POST` with `Stream-Closed: true` (with or without body) closes the stream
- **Append rejection**: Appends to closed streams return `409 Conflict` with `Stream-Closed: true`
- **Reader notification**: `HEAD`, `GET`, long-poll, and SSE responses include closure state

### Reference Server

**Store layer** (`store.ts`, `file-store.ts`):
- `closeStream()` and `closeStreamWithProducer()` methods
- Append rejection when stream is closed
- `closedBy` tracking for idempotent producer duplicate detection
- Long-polls return immediately when stream is closed

**HTTP handlers** (`server.ts`):
- Create, append, HEAD, GET handlers support `Stream-Closed` header
- Long-poll returns immediately with `204 + Stream-Closed: true`
- SSE includes `streamClosed: true` in final control event (without `streamCursor`)

### TypeScript Client

**Writer API**:
- `DurableStream.close(opts?)` - Close stream with optional final message
- `CreateOptions.closed` - Create stream in closed state
- `IdempotentProducer.close(finalMessage?)` - Flush pending + close stream
- `IdempotentProducer.detach()` - Stop producer without closing (renamed from old `close()`)

**Reader API**:
- `streamClosed: boolean` on `HeadResult`, `StreamResponse`, and chunk metadata
- Live modes stop polling/reconnecting when `streamClosed` is true

**Error handling**:
- `StreamClosedError` thrown on 409 with `Stream-Closed` header

### Conformance Tests

- **Server**: ~26 new tests covering create, close, HEAD, GET, long-poll, SSE, and idempotent producer scenarios
- **Client**: 16 new YAML tests covering writer operations, reader catch-up, live modes, and state transitions

### Breaking Changes

- `IdempotentProducer.close()` now closes the stream. Use `detach()` to stop the producer without closing.

Closes #215

---

### Test plan

- [x] All existing tests pass (no regressions)
- [x] Server conformance tests pass for in-memory and file-backed stores
- [x] Client conformance tests pass for TypeScript client
- [x] Idempotent producer close is retry-safe
- [x] Append to closed stream throws `StreamClosedError` with correct code


samwillis:
Thoughts on review:

> 1.1 EOF signaling and Stream-Up-To-Date interaction

I think it's self evident that a chunk with `stream-closed: true` is the end of the stream and therefore up-to-date. It would be very odd for a client to then move to live mode. This feels like a tiny nit.

> 1.2 Create-with-close: Content-Type expectations

I think the protocol is fine, it makes it clear that you have to create (even in closed state) a stream with a content type as it must have one. Closing an existing stream with no content type header makes sense as it provide a route for an admin task to do this without the metadata of what the stream holds.

> 1.3 EOF discovery flow vs. caching behavior

Review agent is misunderstanding. The stream-close can be on either the last chunk or the next empty offset, it depends on when the response was cached. There is nothing to fix here.

> 2.1 Close-only idempotency vs. in-flight writes

This is an application concern, the server handle responses on a first come first served bases.

> 2.3 Long-poll behavior on closed streams

This is in the spec

> 3.1 Cursor omission on closed streams

I will add the test

> 4) Missing Tests (Unit + Conformance)

I will add them

samwillis:
Cursors / Opus4.5 assessment:

=====

Based on my review of the conformance tests against PROTOCOL.md, here's the analysis:

## Tests We Have

**4.1 Server Tests (all present):**
- ✅ `close-stream-empty-post` - Close-only POST
- ✅ `close-idempotent` - Idempotent close returns 204
- ✅ `append-to-closed-stream-409` - Append after close rejected
- ✅ `close-with-final-append` + `append-and-close-to-closed-stream-409` - Append+close atomicity
- ✅ `create-closed-stream` + `create-closed-stream-with-body` - Create closed stream

**4.2 Read Mode Tests (all present):**
- ✅ `read-closed-stream-empty-body-eof` - Catch-up EOF (200 OK, empty body, Stream-Closed)
- ✅ `longpoll-closed-stream-immediate` + `longpoll-closed-returns-204-with-header` - Long-poll EOF
- ✅ `sse-closed-stream-connection-closes` - SSE EOF and connection close
- ✅ `sse-closed-stream-no-cursor` - SSE cursor omitted when closed

**3.1 Cursor Behavior:**
- ✅ `sse-closed-stream-no-cursor` - Verifies SSE `streamCursor` is omitted when closed
- ✅ Existing tests verify cursor IS present on open streams (`should generate Stream-Cursor header on long-poll responses`)
- ✅ Client code uses optional types (`streamCursor?: string`) and conditionally updates cursor

## Assessment

Our implementation and tests are **compliant with PROTOCOL.md**:

1. **Cursor omission**: The spec says servers "MAY omit" cursor when closed (not "MUST omit"). Our server includes it for HTTP but omits it for SSE, both are valid. The SSE test explicitly verifies omission.

2. **Client tolerance**: The client already handles missing cursor - it's defined as optional (`streamCursor?: string`) and the code only updates cursor if present: `if (controlEvent.streamCursor) { this.#cursor = controlEvent.streamCursor }`

3. **All reviewer's suggested tests exist** - Every test in section 4.1 and 4.2 is already implemented.

No additional tests are needed - the review's concerns are already addressed by our existing test coverage.

KyleAMathews:
## Research: Protocol Design Comparison & Conformance Test Gaps

I did a deep comparison of Durable Streams' design choices against Kafka, Kinesis, Pulsar, NATS JetStream, Redis Streams, EventStoreDB, SSE, and gRPC streaming. Summary of findings below, plus a list of conformance tests worth adding.

---

### Design Comparison Highlights

**Stream Closure** — The `Stream-Closed: true` header fills a genuine gap. No comparable system provides a clean, durable, out-of-band EOF signal for HTTP streams. It's better than Redis/SSE's in-band sentinel values (which pollute the data stream), more nuanced than EventStoreDB's hard delete (which destroys data), and addresses something Kafka simply doesn't have. The atomic "append final message + close" is particularly distinctive — I couldn't find an equivalent in other systems.

**Idempotent Producers** — The `(producerId, epoch, seq)` design is a strong adaptation of Kafka's proven pattern with HTTP-native improvements:
- Client-supplied Producer-Id eliminates Kafka's `InitProducerId` round-trip
- Client-declared epochs with server validation is simpler than Kafka's transaction coordinator
- `2^53-1` integer limits are wise for JS interop (Kafka wraps at `int32`/`int16`)
- The auto-claim flow (detect 403, retry with epoch+1) for ephemeral producers is a pragmatic pattern not found in Kafka or Pulsar

**409 for Closed Streams** — Correct per RFC 9110 ("conflict with the current state of the target resource"). One observation: 409 is used for multiple conditions (content-type mismatch, sequence regression, stream closed, producer sequence gap), all disambiguated by different response headers. This works but could benefit from a note in the spec about how clients should distinguish them.

**HTTP Headers** — Clean design. `Stream-` prefix provides good namespacing, avoids deprecated `X-` prefix. The `Producer-Expected-Seq`/`Producer-Received-Seq` diagnostic headers on 409 are better debugging than Kafka provides.

---

### Suggested Conformance Tests

Organized by priority. These are informed by edge cases that have historically caused issues in comparable systems.

#### Critical (data loss / hanging connections)

| Test | What it verifies |
|------|-----------------|
| `concurrent-append-and-close` | Two concurrent requests (one append, one close) serialize correctly — no data loss, consistent final state |
| `close-during-active-long-poll` | Close wakes a waiting long-poll immediately (spec: "MUST NOT wait for the long-poll timeout") |
| `close-during-active-sse` | Close pushes `streamClosed: true` control event to all active SSE connections |
| `catch-up-closure-discovery` | Two-step flow: first read returns data without `Stream-Closed` (cached), second read at tail discovers closure |

#### High (protocol ambiguity / silent bugs)

| Test | What it verifies |
|------|-----------------|
| `atomic-close-dedup-with-producer` | Retry of append-and-close with same `(pid, epoch, seq)` returns 204 and stream has exactly one copy of data |
| `producer-state-survives-close` | After close, stale-epoch producer still gets 403 (not 409/STREAM_CLOSED) — error code matters for client logic |
| `empty-post-without-stream-closed-header` | POST with empty body but no `Stream-Closed: true` returns 400, not treated as close |
| `content-type-on-close-only-request` | Close with empty body + mismatched Content-Type succeeds (spec: "servers MUST NOT reject based on Content-Type" when body is empty) |
| `json-empty-array-close` | `[]` + `Stream-Closed: true` — is this rejected (empty array per Section 7.1) or accepted as close-only? Needs clarification |
| `close-with-different-body-dedup` | Producer sends close+body "A", retries same seq with body "B" — should dedup to "A" |

#### Medium (edge case correctness)

| Test | What it verifies |
|------|-----------------|
| `concurrent-close-and-close` | Two simultaneous close requests both succeed (idempotent), stream closed exactly once |
| `concurrent-append-and-close-with-body` | Race between regular append and append-and-close — both data items present or append rejected |
| `multiple-producers-interleaved` | Two different Producer-Ids writing concurrently — independent sequence spaces don't interfere |
| `producer-epoch-max-value` | Epoch at `2^53-1` — accepted? What happens at `epoch+1`? |
| `producer-seq-max-value` | Same boundary test for Producer-Seq |
| `read-offset-beyond-closed-tail` | Offset past the final data returns empty body + `Stream-Closed: true` + `Stream-Up-To-Date: true` |
| `offset-now-on-closed-stream` | `offset=now` on closed stream returns closure signal with tail offset |
| `long-poll-on-closed-empty-stream` | Create stream, close immediately, long-poll — returns 204 immediately |
| `head-closed-stream-tail-offset` | HEAD after close-with-append returns offset *after* the final data, not before it |
| `stream-closed-header-non-true-values` | `Stream-Closed: false` / `Stream-Closed: yes` / `Stream-Closed: 1` — behavior should be specified |
| `put-idempotent-create-closed-mismatch` | PUT closed stream, then PUT same URL without `Stream-Closed` — 200 or 409? |
| `close-nonexistent-stream` | POST with `Stream-Closed: true` to nonexistent URL returns 404 |
| `producer-state-after-delete-recreate` | Producer writes, stream deleted, stream recreated — old producer state must not carry forward |

#### Lower Priority (robustness / defense-in-depth)

| Test | What it verifies |
|------|-----------------|
| `etag-on-closed-stream` | ETag + `If-None-Match` → 304 works correctly on closed streams |
| `cache-control-on-eof-response` | Cache-Control on the empty-body closure response (this is immutable — could be aggressively cached) |
| `create-closed-with-ttl` | Closed stream + TTL — both apply, stream expires normally |
| `delete-closed-stream` | Deletion overrides closure — returns 404 after, not `Stream-Closed` |
| `head-during-active-sse` | HEAD doesn't interfere with active SSE connections |
| `413-does-not-advance-producer-state` | Payload Too Large rejection doesn't consume a sequence number |
| `producer-id-special-characters` | Spaces, colons, Unicode, very long strings in Producer-Id |
| `stream-seq-combined-with-producer-headers` | Both validations apply independently when both header sets are present |
| `offset-value-resembling-sentinel` | Server must never generate `-1` or `now` as actual offset values |
| `multiple-readers-during-close` | Reader at offset 0 catches up normally; reader at tail gets immediate closure signal |
| `json-batch-append-and-close` | JSON array flattening + atomic close combined |
| `sse-reconnect-after-close` | Client reconnects after receiving closure — gets closure signal again, doesn't hang |

samwillis:
First round of fixes from review:

========


Here's the complete status of everything from the PR review on #220:

## Documentation Gaps
| Item | Status |
|------|--------|
| Client README needs stream closure docs | ✅ Done |

## Confirmed Bugs (Must Fix Before Merge)

| Bug | Status |
|-----|--------|
| 1. IdempotentProducer.close() retry drops finalMessage and skips re-flush | ✅ Fixed |
| 2. Auto-claim in #doClose uses seq 1 instead of 0 | ✅ Fixed |
| 3. SSE streamClosed doesn't imply upToDate | ✅ Fixed |
| 4. ETag doesn't vary with closure status | ✅ Fixed |
| 5. utils.ts uses case-sensitive header check | ✅ Fixed |
| 6. 409 on closed stream should include tail offset | ✅ Fixed |

## Protocol Spec Clarifications

| Clarification | Status |
|---------------|--------|
| Stream-Closed header value rules (only `true` is valid) | ✅ Added |
| PUT idempotency includes closure status matching | ✅ Added |
| Error precedence (closed → content-type → sequence) | ✅ Added |
| offset=now on closed stream behavior | ✅ Added |
| ETag must vary with closure status | ✅ Added |

## Conformance Tests

### Added Tests (14 total)

**Client conformance tests (7):**
- `atomic-close-dedup-with-producer`
- `close-nonexistent-stream`
- `read-offset-beyond-closed-tail`
- `head-closed-stream-returns-tail-offset`
- `delete-closed-stream`
- `create-closed-with-ttl`
- `close-with-body-retry-different-body`

**Server conformance tests (7):**
- `409-includes-stream-offset`
- `close-nonexistent-stream-404`
- `offset-now-on-closed-stream`
- `producer-state-survives-close`
- `close-with-different-body-dedup`
- `empty-post-without-stream-closed-400`
- `delete-closed-stream`

### Missing Tests (from review suggestions)

**Critical (concurrency tests - complex to implement):**
- `concurrent-append-and-close` - Two concurrent requests serialize correctly
- `close-during-active-long-poll` - Close wakes waiting long-poll immediately
- `close-during-active-sse` - Close pushes streamClosed to active SSE connections
- `catch-up-closure-discovery` - Two-step closure discovery flow

**High priority:**
- `json-empty-array-close` - `[]` + Stream-Closed behavior

**Medium priority:**
- `concurrent-close-and-close` - Two simultaneous closes
- `concurrent-append-and-close-with-body` - Race between append and append-and-close
- `multiple-producers-interleaved` - Two Producer-Ids writing concurrently
- `producer-epoch-max-value` - Epoch at 2^53-1 boundary
- `producer-seq-max-value` - Seq at 2^53-1 boundary
- `long-poll-on-closed-empty-stream` - Create, close immediately, long-poll
- `stream-closed-header-non-true-values` - Stream-Closed: false/yes/1 behavior
- `put-idempotent-create-closed-mismatch` - PUT closed, then PUT without closed
- `producer-state-after-delete-recreate` - Old producer state doesn't carry forward

**Lower priority:**
- `etag-on-closed-stream` - ETag + If-None-Match → 304 works
- `cache-control-on-eof-response` - Cache-Control on closure response
- `head-during-active-sse` - HEAD doesn't interfere with SSE
- `413-does-not-advance-producer-state` - Payload Too Large doesn't consume seq
- `producer-id-special-characters` - Spaces, colons, Unicode in Producer-Id
- `stream-seq-combined-with-producer-headers` - Both validations apply
- `offset-value-resembling-sentinel` - Server never generates -1 or "now"
- `multiple-readers-during-close` - Reader at 0 catches up, reader at tail gets closure
- `json-batch-append-and-close` - JSON array flattening + close
- `sse-reconnect-after-close` - Reconnect after closure gets signal again

## Other Items (Not Bugs)

| Item | Status |
|------|--------|
| Create-closed race window in file store | ❌ Not addressed (noted as dev server only concern) |

---

## Summary

**Completed:** All 6 bugs, all 5 spec clarifications, documentation, 14 conformance tests

**Not completed:** ~20 additional conformance tests (mostly concurrency tests that are complex to implement in the current test framework)

PR #232, opened 2026-01-29
feat: TanStack AI chat example with durable proxy integration

balegas:
## Summary

Adds a TanStack AI chat example app that routes streaming responses through the durable proxy for resumable AI streaming. Also rewrites the `@durable-streams/proxy` TanStack transport to match the actual `@tanstack/ai-client` `ConnectionAdapter` interface.

**Status: WIP** — proxy forwarding works, but stream closure errors and stream resume are not yet working correctly.

### Approach

The integration has three parts:

1. **Fixed TanStack transport** (`packages/proxy/src/transports/tanstack.ts`): The existing adapter used a custom `ConnectionAdapter` interface (`connect(options) → Promise<{stream, headers, status}>`) that didn't match TanStack's actual interface (`async *connect(messages, data?, abortSignal?) → AsyncIterable<StreamChunk>`). Rewrote it as an async generator that uses `durableFetch` for the two-phase POST→GET flow, then parses SSE from the response body into individual JSON chunks.

2. **Durable proxy startup script** (`examples/chat-tanstack/server/durable-proxy.ts`): Starts both the DS storage server (port 4437) and the proxy server (port 4440) with an allowlist for `localhost:3002`.

3. **Chat app wiring** (`examples/chat-tanstack/src/Chat.tsx`): Replaced `fetchServerSentEvents('/api/chat')` with `createDurableAdapter(...)`. Stream keys are scoped per conversation turn (`{conversationId}-turn-{messageCount}`).

**Architecture flow:**
```
Browser → Durable Proxy (:4440) → Chat Server (:3002) → Anthropic API
                ↕
     DS Storage Server (:4437)
```

### Known Issues

- **Stream closure fails with error**: After the upstream finishes, the stream does not close cleanly. Needs investigation into how the proxy signals stream completion and how the client detects it.
- **Stream resume not working**: On page refresh, `durableFetch` should detect stored credentials and reconnect via GET with the last known offset, but this path isn't functioning correctly yet.

### Non-goals

- No changes to the Vercel AI transport (only TanStack)
- No new tests yet (WIP)
- No production deployment config

### Trade-offs

- Used `as any` cast for the connection adapter type since `@durable-streams/proxy` doesn't depend on `@tanstack/ai` — avoids adding a peer dependency for now
- The example sends raw `UIMessage` format to the server (not `ModelMessage`), which works because the server's `convertMessagesToAnthropic` handles both formats

## Verification

```bash
cd examples/chat-tanstack
pnpm install
cp .env.example .env  # then set ANTHROPIC_API_KEY
pnpm dev              # starts all 3 servers + Vite client
```

## Files changed

| File | Description |
|------|-------------|
| `packages/proxy/src/transports/tanstack.ts` | Rewrote adapter to match TanStack's `ConnectionAdapter` async generator interface; added SSE parsing |
| `examples/chat-tanstack/package.json` | New workspace example with deps on proxy, server, TanStack AI |
| `examples/chat-tanstack/server/durable-proxy.ts` | Script that starts DS server + proxy server |
| `examples/chat-tanstack/server/index.ts` | Chat server (copied from original app, unchanged) |
| `examples/chat-tanstack/src/Chat.tsx` | Uses `createDurableAdapter` instead of `fetchServerSentEvents` |
| `examples/chat-tanstack/src/*` | Other UI files copied from original app |
| `pnpm-lock.yaml` | Updated lockfile for new example deps |

### DIFF
--- a/examples/chat-tanstack/.env.example
+++ b/examples/chat-tanstack/.env.example
@@ -1,0 +1,11 @@
+# Required: Your Anthropic API key
+ANTHROPIC_API_KEY=your-api-key-here
+
+# Optional: Override the default model (claude-sonnet-4-20250514)
+# MODEL=claude-sonnet-4-20250514
+
+# Optional: Durable proxy settings (defaults shown)
+# PROXY_SECRET=dev-secret
+# VITE_PROXY_URL=http://localhost:4440/v1/proxy
+# VITE_PROXY_SECRET=dev-secret
+# VITE_UPSTREAM_URL=http://localhost:3002/api/chat
--- a/examples/chat-tanstack/.gitignore
+++ b/examples/chat-tanstack/.gitignore
@@ -1,0 +1,22 @@
+# Dependencies
+node_modules/
+
+# Build
+dist/
+
+# Environment
+.env
+.env.local
+.env.*.local
+
+# IDE
+.idea/
+.vscode/
+*.swp
+*.swo
+
+# OS
+.DS_Store
+
+# Logs
+npm-debug.log*
--- a/examples/chat-tanstack/index.html
+++ b/examples/chat-tanstack/index.html
@@ -1,0 +1,12 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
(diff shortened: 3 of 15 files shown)


balegas:
I agree... too much vibes, not enough thinking here. Need to take a step back.

PR #236, opened 2026-01-29
Support +json suffix content types in JSON Mode

KyleAMathews:
## Summary

Extends JSON Mode to recognize content types with a `+json` suffix (e.g., `application/vnd.api+json`, `application/ld+json`) per RFC 6839, in addition to `application/json`. This enables proper JSON array handling for APIs using structured media types.

## Behavior Change Note

**For existing streams with `+json` content types:** If you previously created streams with content types like `application/vnd.api+json`, they would have been treated as byte-stream mode. With this change, they will now be interpreted as JSON mode (message boundaries, array wrapping on reads, array flattening on append). This may affect:
- How data is read (JSON array vs raw concatenation)
- SSE encoding (text vs base64)

This is intentional—`+json` suffix semantically indicates JSON-parseable content per RFC 6839.

## Approach

**Strategy**: Add a shared `isJsonContentType()` helper across all implementations that normalizes and checks content types consistently.

The helper:
1. Extracts the media type (before any semicolon for charset params)
2. Converts to lowercase for case-insensitive matching
3. Returns true for exact `application/json` OR any `+json` suffix

**Key invariants**:
- All existing `application/json` behavior preserved
- Content type matching is case-insensitive per HTTP spec
- Parameters (like `charset=utf-8`) are properly ignored per RFC 8259
- Consistent behavior across all 10 client implementations

**Non-goals**:
- Other structured suffixes (`+xml`, `+yaml`) are explicitly NOT treated as JSON
- `text/json` is NOT treated as JSON mode (only `application/` prefix)

## Verification

```bash
# Run all conformance tests
pnpm test:run

# Run specific client tests
pnpm test:run -- --client python
pnpm test:run -- --client go
```

## Files Changed

**Protocol & Docs**
- `PROTOCOL.md` - Document JSON Mode content type matching with RFC references

**TypeScript Client** (`packages/client/`)
- `constants.ts` - Add `isJsonContentType()` and `normalizeContentType()` helpers
- `stream.ts`, `stream-api.ts`, `idempotent-producer.ts`, `error.ts` - Use new helpers

**Server** (`packages/server/`)
- `server.ts`, `store.ts`, `file-store.ts` - Use `isJsonContentType()` for JSON mode detection

**Go Server** (`packages/caddy-plugin/`)
- `handler.go`, `store/store.go` - Add `IsJSONContentType()` helper
- `store/memory_store.go` - Use shared helper (removed duplicate)

**Client SDKs** (TypeScript, Python, Go, Java, PHP, Rust, .NET)
- Each adds `isJsonContentType()` helper with identical logic
- Java uses `Locale.ROOT` for locale-safe case conversion
- Conformance adapters updated to use the helper for test verification

**Tests**
- `json-content-type-variants.yaml` - Comprehensive conformance tests for +json handling

### DIFF
--- a/PROTOCOL.md
+++ b/PROTOCOL.md
@@ -789,7 +789,7 @@ The `+json` structured syntax suffix is defined by [RFC 6839](https://www.rfc-ed
 
 #### Encoding
 
-Per [RFC 8259](https://www.rfc-editor.org/rfc/rfc8259), JSON text exchanged between systems **MUST** be encoded using UTF-8. Servers **SHOULD** assume UTF-8 encoding for JSON streams unless a charset parameter explicitly specifies otherwise.
+Per [RFC 8259](https://www.rfc-editor.org/rfc/rfc8259), JSON mode payloads **MUST** be UTF-8 encoded. Receivers **MUST** interpret JSON mode payloads as UTF-8. If a `charset` parameter is present, it **MUST** be ignored (RFC 8259 specifies that `application/json` has no charset parameter).
 
 #### Message Boundaries
 
--- a/packages/caddy-plugin/store/memory_store.go
+++ b/packages/caddy-plugin/store/memory_store.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
-	"strings"
 	"sync"
 	"time"
 )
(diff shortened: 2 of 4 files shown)


samwillis (approved):
On the whole this looks good, my one hesitation is that many content types that use json have a very specific structure, often with an object at the top level. Our json mode if an array at the top velocity from the appearance on the read path. Many of these formats could be appended once but not continually appended too.

Approving ahead of any fixes from addressing the chatgpt review.

PR #246, opened 2026-02-05
feat: webhook subscriptions — protocol, server, conformance tests

KyleAMathews:
Add push-based webhook delivery to Durable Streams. Subscriptions map glob patterns to webhook URLs; when matching streams receive events, the server POSTs a signed notification, spawning a consumer instance that tracks offsets across streams and reports progress via a scoped callback API.

## Approach

The webhook system is a state machine with three consumer states: **IDLE → WAKING → LIVE → IDLE**. Key protocol mechanisms:

- **Epoch fencing** — monotonically increasing counter prevents zombie consumers from interfering with current wake cycles
- **Wake ID claims** — unique per-wake identifier ensures exactly-once processing even with retry-induced duplicate deliveries
- **Callback token rotation** — HMAC-signed tokens with 1-hour TTL, rotated on every successful callback response
- **Webhook signatures** — `HMAC-SHA256(secret, "<timestamp>.<body>")` with timing-safe comparison

The implementation spans 6 new server modules (`webhook-{types,glob,crypto,store,manager,routes}.ts`) integrated via an opt-in `webhooks: true` server option. The testing infrastructure includes a fluent DSL and TLA+-inspired invariant checkers.

### Key invariants

These are checked automatically on every test run:

| ID | Property | What it catches |
|----|----------|-----------------|
| S1 | Epoch monotonicity | Stale wake cycles reusing old epochs |
| S2 | Wake ID uniqueness | Duplicate wake IDs across notifications |
| S3 | Single claim | Wake ID claimed more than once |
| S5 | Token rotation | Response token same as request token |
| S8 | Signature presence | Missing or invalid webhook signatures |
| L1 | Pending work → wake | Idle consumer not woken after append |
| L3 | Done + pending → re-wake | Consumer goes idle despite pending events |

### Non-goals

- **Production hardening** — no SSRF protection, request body size limits, or structured logging (this is the dev server)
- **Persistent consumer state** — consumers are in-memory only; the file-backed store doesn't persist webhook state
- **Multi-server coordination** — no distributed locking or consumer handoff

### Trade-offs

- **Fire-and-forget webhook delivery** rather than awaiting — `onStreamAppend` is called synchronously from the append path, so blocking would add latency to every write. The `deliverWebhook` method handles its own error recovery internally.
- **History-based invariant checking** over individual assertions — records all observable events into an ordered trace and verifies global properties post-hoc. More powerful (catches ordering violations that point assertions miss) at the cost of slightly harder debugging when a check fails.
- **Epoch !== check** (not just <) — the spec says "must match current epoch" so we reject both stale and future epochs, even though future epochs are theoretically impossible since the server controls epoch values.

## Verification

```bash
# Run all 565 server tests (includes 35 webhook-specific tests)
pnpm vitest run --project server

# Build affected packages
pnpm -r --filter '@durable-streams/server' build
cd packages/server-conformance-tests && pnpm build
```

### Specification & docs
- **`PROTOCOL.md`** — New Section 6: Webhook Subscriptions protocol spec
- **`docs/webhook-spec.md`** — Standalone feature specification (source of truth)
- **`docs/webhooks-rfc.md`** — Design RFC with rationale and examples

### Server implementation (`packages/server/`)
- **`webhook-types.ts`** — Type definitions (Subscription, ConsumerInstance, CallbackRequest/Response)
- **`webhook-glob.ts`** — Glob pattern matching for stream paths (`*`, `**`)
- **`webhook-crypto.ts`** — HMAC-SHA256 signatures and callback token generation/validation
- **`webhook-store.ts`** — In-memory consumer/subscription state with index management
- **`webhook-manager.ts`** — Orchestration: wake cycles, retry scheduling, callback processing, liveness timeouts
- **`webhook-routes.ts`** — HTTP handlers for subscription CRUD and callback API
- **`server.ts`** — Integration: opt-in `webhooks: true`, lifecycle hooks
- **`types.ts`** — Added `webhooks` option to `TestServerOptions`
- **`index.ts`** — Re-exports for webhook types

### Conformance tests (`packages/server-conformance-tests/`)
- **`webhook-dsl.ts`** — Fluent test builder, webhook receiver, history recorder, invariant checkers (S1-S8, L1-L3), ENABLED predicates for property-based testing
- **`index.ts`** — 35 webhook test cases including property-based random action sequences

### DIFF
--- a/PROTOCOL.md
+++ b/PROTOCOL.md
@@ -717,9 +717,11 @@ data: {"streamNextOffset":"123456_789","streamCursor":"abc"}
 
 ## 6. Webhook Subscriptions
 
-Webhook subscriptions enable serverless functions and AI agents to react to stream events without maintaining persistent connections. Unlike traditional webhooks that deliver data inline, Durable Streams webhooks are **wake-up signals** — the notification tells the consumer _which_ streams have new events, and the consumer reads the actual data using the standard HTTP protocol (Sections 3–4).
+Webhook subscriptions enable serverless functions and AI agents to react to stream events without maintaining persistent connections.
 
-The lifecycle works like this: a subscription maps a glob pattern to a webhook URL. When a matching stream receives events, the server POSTs a notification to wake the consumer. The consumer claims the wake, reads events through normal Durable Streams reads, acknowledges progress via a callback API, and signals completion when done. The server tracks offsets across all subscribed streams so the consumer resumes exactly where it left off on the next wake.
+**Important:** Durable Streams webhooks do **not** deliver event data inline. They are purely **wake-up signals** that manage a consumer's lifecycle — waking it when there is work, and tracking that it has acknowledged processing. The consumer reads the actual event data itself using the standard HTTP read protocol (Sections 3–4).
+
+The lifecycle: a subscription maps a glob pattern to a webhook URL. When a matching stream receives new events, the server POSTs a lightweight notification to wake the consumer. The consumer claims the wake via a callback API, reads events through normal Durable Streams reads, acknowledges progress (so the server knows where it left off), and signals completion when done. On the next wake, the consumer resumes exactly where it left off — no data is lost or re-delivered.
 
 ### 6.1. Subscription Model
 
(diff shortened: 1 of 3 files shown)


samwillis (commented):
I've had a good read of the protocol spec. I like it, it's not too big and covers the core use case we are going for. This is going to work very well.

I think we can likely tighten up the overview a little so it's easer for someone to understand when the come with some preconceived ideas (`appended data => web hook`, rather than `consumer lifecycle => web hook`)

packages/server-conformance-tests/src/webhook-dsl.ts
```diff
@@ -0,0 +1,1450 @@
...
+        `S2: Duplicate wake_id ${event.wake_id} for consumer ${event.consumer_id}`
+      ).toBe(false)
+      wakeIds.add(event.wake_id)
+      wakeIdsByConsumer.set(event.consumer_id, wakeIds)
+    }
+  }
+}
+
+/** S3: Claiming the current wake_id is idempotent; non-matching wake_ids are rejected. */
+function checkSingleClaim(_history: Array<HistoryEvent>): void {
```
kevin-dp: Why do we have this function if it does nothing?

PR #254, opened 2026-02-11
Refactor StreamResponseImpl into an explicit state machine

kevin-dp:
This PR replaces ~10 mutable fields on `StreamResponseImpl` with a single immutable `#syncState: StreamResponseState` object. State transitions return new objects — no mutation.

### State hierarchy

```
StreamResponseState (abstract)
├── LongPollState       shouldUseSse() → false
├── SSEState            shouldUseSse() → true  (+resilience tracking)
└── PausedState         wraps inner state, delegates getters
```

### What changed

- `#offset`, `#cursor`, `#upToDate`, `#streamClosed`, `#sseFallbackToLongPoll`, `#consecutiveShortSSEConnections`, `#lastSSEConnectionStartTime` → single `#syncState`
- `#justResumedFromPause` one-shot flag → local variable in pull loop
- SSE fallback encoded as type (`SSEState` vs `LongPollState`) instead of boolean
- `handleConnectionEnd()` is a pure function — takes `now` as parameter, returns action + new state

### Tests

48 new pure unit tests for the state machine (synchronous, no mocking). All 316 existing tests pass.

### DIFF
--- a/.changeset/giant-trams-stay.md
+++ b/.changeset/giant-trams-stay.md
@@ -1,0 +1,5 @@
+---
+"@durable-streams/client": patch
+---
+
+Extract pure helper functions from StreamResponseImpl class to module-level functions
--- a/packages/client/src/response.ts
+++ b/packages/client/src/response.ts
@@ -394,158 +394,6 @@ export class StreamResponseImpl<
   }
 
   /**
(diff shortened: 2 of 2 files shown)


KyleAMathews (approved):
:shipit:

PR #293, opened 2026-03-19
Port production hardening features from Electric's TypeScript client

### LINKED ISSUE #265: Port production hardening features from Electric's TypeScript client
## Summary

The Electric SQL TypeScript client (`@electric-sql/client`) has accumulated significant production hardening from real-world deployments — connection resilience, proxy/CDN workarounds, error recovery, and defensive detection of pathological states. The durable-streams TypeScript client should port these features to reach the same level of robustness.

This issue tracks the full set of features to bring over, ordered by priority.

---

## Context

The durable-streams client's state machine (on `kevin/state-machine-refactor`) was modeled after Electric's `ShapeStream` state machine. Both use the same core pattern: immutable class hierarchy, `PausedState` wrapper, SSE-to-long-poll fallback. But Electric has several hardening layers the durable-streams client currently lacks.

Reference implementation: [`electric/packages/typescript-client/src/`](https://github.com/electric-sql/electric/tree/main/packages/typescript-client/src)

---

### 1. Multi-Reason PauseLock

**Priority:** P0 — foundation for other features
**Electric source:** `pause-lock.ts` (~112 lines)
**Effort:** Small

**Problem:** The current pause/resume in `response.ts` uses a manual `#state` string enum (`active` / `pause-requested` / `paused`) with hand-rolled promise plumbing. This only supports a single reason for pausing (visibility change). Future features (snapshots, explicit user pause, system wake) need to coexist without interfering.

**Solution:** Port Electric's `PauseLock` — a `Set<string>`-based counting lock where multiple named reasons can hold the pause. The stream pauses on the `0 → 1` transition (first reason acquired) and resumes on the `1 → 0` transition (last reason released).

```
acquire('visibility')  → paused (1 holder)
acquire('snapshot-1')  → still paused (2 holders)
release('visibility')  → still paused (1 holder)
release('snapshot-1')  → resumed (0 holders)
```

Key behaviors:
- `acquire()` fires `onAcquired` callback only on `0 → 1` transition
- `release()` fires `onReleased` callback only on `1 → 0` transition
- Duplicate `acquire` / orphan `release` log warnings (idempotent, no crash)
- `releaseAllMatching(prefix)` for batch cleanup (e.g., clearing all `snapshot-*` reasons)

---

### 2. System Wake Detection

**Priority:** P0 — prevents stale connections after sleep
**Electric source:** `client.ts` `#subscribeToWakeDetection()` (~50 lines)
**Effort:** Small

**Problem:** When a laptop sleeps and wakes, the in-flight HTTP connection (long-poll or SSE) may be dead but the client doesn't know. It can sit waiting on a stale socket indefinitely. Browsers handle this via `document.visibilitychange`, but Node.js and Bun have no equivalent.

**Solution:** In non-browser environments, run a `setInterval` timer at 2-second intervals. If the elapsed wall-clock time between ticks exceeds 6 seconds (2s interval + 4s threshold), the system was sleeping. On detection:

1. Abort the in-flight request with a `SYSTEM_WAKE` reason
2. The request loop catches this and immediately restarts from the current offset
3. No state reset needed — just reconnect

Implementation details:
- Skip in browsers (visibility API handles it)
- Call `timer.unref()` so the timer doesn't keep Node/Bun alive
- Clean up on stream teardown

**Depends on:** PauseLock (uses its abort infrastructure)

---

### 3. Fast-Loop Detection

**Priority:** P1 — safety net against proxy misconfiguration
**Electric source:** `client.ts` `#checkFastLoop()` (~70 lines)
**Effort:** Small–Medium

**Problem:** A misconfigured proxy or CDN can return empty 200 responses without advancing the stream offset. The client will re-request from the same offset immediately, creating a tight infinite loop that burns CPU and network bandwidth with no visible error.

**Solution:** Track recent non-live requests in a sliding time window. If multiple requests arrive at the same offset within the window, the client is stuck in a loop.

Algorithm:
- Maintain an array of `{ timestamp, offset }` entries
- Before each non-live request, prune entries older than 500ms
- If ≥ 5 entries share the same offset within the window, trigger detection
- Only checked during syncing (not live) — live requests clear the counters

Escalation ladder:
1. **Detection 1:** Log warning, clear client caches, reset stream to offset `-1`
2. **Detections 2–4:** Exponential backoff with full jitter (base 100ms, cap 5s)
3. **Detection 5:** Fatal error — throw with a descriptive message about proxy misconfiguration

The key insight is that the sliding window is time-based but the trigger condition checks *offset identity* — normal rapid syncing (where each response advances the offset) never trips the detector.

---

### 4. ErrorState + Resumable `onError`

**Priority:** P1 — biggest reliability gap today
**Electric source:** `shape-stream-state.ts` `ErrorState` (~65 lines) + wiring in `client.ts`
**Effort:** Medium

**Problem:** Currently, errors during the active long-poll/SSE loop call `#markError()` and **permanently close the stream**. A transient 503 during an active session kills it with no recovery path. The `onError` handler only fires on the initial `stream()` call, not during ongoing polling.

**Solution:** Add `ErrorState` to the state machine. When an error occurs mid-stream:

1. Wrap current state: `this.#syncState = this.#syncState.toErrorState(error)`
2. `ErrorState` preserves the full sync position (offset, cursor, transport mode)
3. Call the user's `onError` handler
4. If `onError` returns an object (even `{}`), call `errorState.retry()` which unwraps back to the previous state and resumes from the same position
5. If `onError` returns `void`, propagate the error (stream closes)

This turns fatal mid-stream errors into recoverable ones. The user can refresh auth tokens, wait for a service to recover, or apply any other fix before retrying.

State machine addition:
```
ErrorState {
  previousState: LongPollState | SSEState | PausedState
  error: Error

  retry() → previousState          // unwrap, resume
  pause() → PausedState(this)      // can pause while in error

  // All field getters delegate to previousState
}
```

Nesting prevention: `ErrorState(ErrorState(X))` flattens to `ErrorState(X)` in the constructor.

---

### 5. Fetch Middleware Enhancements

**Priority:** P2 — polish on existing infrastructure
**Electric source:** `fetch.ts` (relevant portions ~120 lines)
**Effort:** Small–Medium

#### 5a. `Retry-After` Header Support

**Problem:** When a server returns 429 (rate limited) or 503 (overloaded), it often includes a `Retry-After` header telling the client how long to wait. The current `createFetchWithBackoff` ignores this header — it may retry too soon (wasting requests and potentially making rate limiting worse) or too late (unnecessary delay).

**Solution:** Parse the `Retry-After` header (supports both delta-seconds and HTTP-date formats) and use it as a **floor** on the backoff delay:

```
waitMs = max(serverRetryAfterMs, clientExponentialBackoffMs)
```

This ensures the client never retries sooner than the server asked, but also doesn't wait longer than its own backoff schedule would dictate.

#### 5b. Response Header Validation

**Problem:** A misconfigured reverse proxy might strip required protocol headers (`durable-stream-offset`, `stream-closed`, `cursor`, etc.) from responses. The client would see a 200 with valid JSON but missing metadata, leading to subtle bugs (wrong offset tracking, missed stream-closed signals).

**Solution:** Add a `createFetchWithResponseHeadersCheck` middleware that validates required protocol headers are present on 2xx responses:

- Always required: `durable-stream-offset`
- When `live=true`: also `cursor`
- When not live: also `durable-stream-schema` (if applicable)

Throw a descriptive `MissingHeadersError` listing which headers are absent, making proxy misconfiguration immediately obvious.

---

### 6. CDN Cache Busting (Stale Response Detection + Retry)

**Priority:** P2 — needed when deploying behind CDNs
**Electric source:** `StaleRetryState` in `shape-stream-state.ts` (~40 lines), detection logic in `client.ts` (~70 lines), `expiredShapesCache` utility
**Effort:** Medium–Large

**Problem:** When a CDN caches stream responses and a stream handle expires (e.g., server restarts, stream is deleted and recreated), the CDN may continue serving the old cached response. The client receives data with an expired handle, re-requests, and gets the same stale cached response in a loop.

This is a real-world issue with any CDN or caching reverse proxy (Cloudflare, Fastly, Varnish, nginx cache, etc.) sitting in front of the durable-streams server.

**Solution:** Multi-layer detection and recovery:

**Layer 1 — Stale response detection:**
When the client requests a stream and the response's handle matches a known-expired handle, the response is stale (served from CDN cache rather than the origin).

**Layer 2 — Cache buster retry:**
Add a new `StaleRetryState` to the state machine that carries a random `cache_buster` query parameter (e.g., `?cache_buster=1709394821-k7f2m9x`). This parameter is meaningless to the server but forces the CDN to treat it as a cache miss.

```
Normal request:  GET /v1/streams/my-stream?offset=0
Stale detected:  GET /v1/streams/my-stream?offset=0&cache_buster=1709394821-k7f2m9x
```

**Layer 3 — Retry limit:**
Track `staleCacheRetryCount`. After 3 failed attempts (CDN still serving stale data despite cache buster), throw a fatal error with a descriptive message about CDN cache misconfiguration. This prevents infinite retry loops.

**Layer 4 — Expired handle tracking:**
Maintain a client-side cache of known-expired stream handles (the `expiredShapesCache` pattern). When a 409 response indicates a handle has expired, record it. On subsequent requests, include the expired handle as a query parameter so the client can detect if the CDN serves it back.

State machine addition:
```
StaleRetryState extends the fetching state hierarchy {
  staleCacheBuster: string        // random cache-busting value
  staleCacheRetryCount: number    // attempts so far

  applyUrlParams(url) → adds ?cache_buster=<random>
}
```

**Server-side requirement:** The server (Caddy plugin) must ignore unknown query parameters so `cache_buster` doesn't cause errors. Verify this is already the case or add it.

**Cross-tab coordination (stretch):** Electric uses `localStorage` to share expired-handle knowledge across browser tabs. Consider whether this is needed for durable-streams or if per-instance tracking is sufficient initially.

---

### 7. Replay Mode (Cached Response Deduplication)

**Priority:** P2 — eliminates duplicate data delivery on page refresh
**Electric source:** `ReplayingState` in `shape-stream-state.ts` (~53 lines), `UpToDateTracker` in `up-to-date-tracker.ts` (~163 lines), wiring in `client.ts`
**Effort:** Medium

**Problem:** When a user refreshes the page, a CDN or caching proxy may serve cached HTTP responses (up to 60s old depending on `s-maxage`). Without replay mode, these cached responses deliver the same data the client already processed in the previous session — duplicate `up-to-date` events and duplicate message batches reach subscribers.

**Solution:** Two-part system — a persistence layer and a state machine state:

**Part 1 — `UpToDateTracker` (persistence):**
A module-level singleton that records the cursor at each `up-to-date` transition in `localStorage`:
- Key: canonical stream URL (stripped of protocol params like offset/cursor/live for stability across requests)
- Value: `{ timestamp, cursor }` — the cursor at the moment of the last up-to-date
- TTL: 60s (matching typical CDN `s-maxage`) — entries older than 60s are ignored since the CDN cache has expired
- Write throttle: 60s — prevents excessive `localStorage` writes while keeping in-memory state current
- LRU eviction at 250 entries
- All `localStorage` I/O wrapped in try/catch (handles SSR, quota exceeded, private browsing)

**Part 2 — `ReplayingState` (state machine):**
Entered when a fresh stream starts and the `UpToDateTracker` has a recent entry for the same stream URL:
- Carries a `replayCursor` (the cursor from the previous session)
- On receiving an `up-to-date` message: compares the current response's cursor to `replayCursor`
  - **Match** (cursor unchanged → CDN served cached response): `suppressBatch = true` — the entire batch is silently discarded
  - **Mismatch** (cursor advanced → fresh data from origin): batch passes through normally
- Either way, transitions to `LiveState` — replay mode is a one-shot gate
- Only applies to long-poll transport (SSE connections are not served from CDN cache)

**Entry conditions:**
1. Stream is not yet up-to-date (still syncing)
2. Current state supports replay (`InitialState`, `SyncingState` — NOT `StaleRetryState`)
3. `UpToDateTracker.shouldEnterReplayMode()` returns a non-null cursor (recent entry exists)

**Adaptation for durable-streams:** The cursor comparison logic maps directly — Electric's `liveCacheBuster` corresponds to durable-streams' `cursor`. The `canonicalShapeKey` function needs to be adapted to strip durable-streams protocol params instead of Electric's.

---

### 8. Chunk Prefetching

**Priority:** P3 — performance optimization for initial sync
**Electric source:** `createFetchWithChunkBuffer` + `PrefetchQueue` in `fetch.ts` (~185 lines), `getNextChunkUrl` helper (~35 lines)
**Effort:** Medium

**Problem:** When a client connects to a stream with a large backlog, it fetches chunks sequentially — request chunk 1, wait for response, process it, request chunk 2, wait, process, etc. Each round-trip adds latency. For a stream with 100 chunks, the total sync time is `100 × (network_RTT + processing_time)` instead of the theoretical minimum where network and processing overlap.

**Solution:** A speculative prefetch layer in the fetch middleware chain. After receiving a chunk response, inspect the response headers to infer the next chunk's URL and fire off the fetch immediately — before the client finishes processing the current chunk.

**Core design — `PrefetchQueue`:**
- A `Map<url, [Promise<Response>, AbortController]>` that tracks in-flight prefetched requests
- Head/tail pointers maintain a linked-list structure over URLs
- **In-order consumption only** — if a consumer requests a URL that's in the queue but not the head, returns `undefined` (cache miss). This prevents out-of-order consumption from corrupting the chain.
- Default: prefetch up to 2 chunks ahead (`maxChunksToPrefetch = 2`)
- Self-chaining: after each prefetch resolves, `#prefetch` calls itself recursively for the next URL
- Per-entry `AbortController` instances (chained to the user's signal via `chainAborter`) so consuming a still-streaming response is not interrupted when the rest of the queue is invalidated

**`getNextChunkUrl` adaptation for durable-streams:**
Electric infers the next URL from `electric-handle` + `electric-offset` response headers. For durable-streams, the equivalent is:
- Read `durable-stream-offset` from the response headers
- Read the cursor (if present)
- Construct next URL with updated `offset` param
- Skip prefetching when: response has `stream-closed: true`, request has `live=true`, or required headers are missing

**Guard rails:**
- Only prefetches GET requests (not POST/idempotent writes)
- Never prefetches live (long-poll) requests — they should stay open for server push
- Errors in prefetched requests are silently swallowed (`.catch(() => {})`) — the real retry happens when the consumer actually needs the chunk, via the backoff layer
- Stale-handle guard: if the response handle matches an expired handle, don't prefetch (would just get another stale response)

**Placement in the fetch middleware chain:**
```
baseFetchClient (raw fetch)
  └── createFetchWithBackoff (retry + backoff)
        └── createFetchWithChunkBuffer  ← PREFETCH LAYER
              └── createFetchWithResponseHeadersCheck (header validation)
                    └── createFetchWithConsumedMessages (eagerly reads body)
```

Prefetching sits inside the header validation layer, meaning prefetched responses are validated when the consumer awaits them (not at prefetch time).

---

## Implementation Plan

| Phase | Features | Notes |
|-------|----------|-------|
| **Phase 1** | PauseLock, System Wake Detection | Foundation; small, self-contained |
| **Phase 2** | Fast-Loop Detection, ErrorState + Resumable `onError` | Core resilience; medium effort |
| **Phase 3** | Fetch middleware (`Retry-After`, header validation), CDN Cache Busting | Production CDN support |
| **Phase 4** | Replay Mode, Chunk Prefetching | Performance + dedup optimizations |

Each phase can be a separate PR. Features within a phase are independent and could also be split into individual PRs.

KyleAMathews:
## Summary

Port 8 production hardening features from Electric's TypeScript client, replacing the 3-state machine with a 7-state Electric-style hierarchy and adding connection resilience, CDN workarounds, error recovery, and performance optimizations. 498 tests (168 new), including a formal SPEC.md with compiler-enforced truth table.

## Root cause

The durable-streams client had no recovery path for mid-stream errors (permanent stream death on transient 503), no protection against CDN stale response loops, no system wake detection (stale connections after laptop sleep), and a single-reason pause mechanism that couldn't coordinate visibility changes with future snapshot operations.

## Approach

Ported Electric's production-hardened patterns, adapted for the durable-streams protocol:

**State machine restructure** — `StreamState > ActiveState > FetchingState > {Initial, Syncing, StaleRetry}` + `LiveState`, `ReplayingState`, `PausedState`, `ErrorState`. Each state owns its URL construction (`applyUrlParams`) and transition logic (`handleResponseMetadata`, `handleMessageBatch`). Immutable — every transition returns a new state.

**Key adaptation**: Stale detection lives in the fast-loop detector, NOT the state machine. The durable-streams protocol allows valid `200 OK` with same offset for at-tail reads (PROTOCOL.md Section 5.6) — offset comparison would false-positive on every catch-up that reaches the tail.

**Fetch middleware chain**: Two separate clients matching Electric's split:
```typescript
chunkFetchClient = consumedBody(headersCheck(chunkBuffer(backoff(base))))
sseFetchClient   = headersCheck(backoff(base))
```
SSE client skips consumed-body (would destroy the event stream) and chunk buffer (not applicable).

## Key invariants

- Every non-live GET request must produce a unique URL (duplicate-URL guard)
- `MissingHeadersError` is non-retryable (stripped headers = broken proxy, retry is pointless)
- `ErrorState(ErrorState(X))` flattens to `ErrorState(X)` (no same-type nesting)
- `PausedState.pause()` returns `this` (idempotent)
- `retry()` returns the exact `previousState` instance (preserves all fields)
- `streamClosed` is monotonically true (once closed, stays closed)
- State machine truth table: 63 cells, all verified by compiler-enforced `Record` with no `Partial`

## Non-goals

- No `expiredShapesCache` — durable-streams URLs are stable, not transient handles
- No `markMustRefetch` / `withHandle` events — no shape handle rotation
- No 204 response semantics — durable-streams uses `Stream-Up-To-Date` header
- No cross-tab `localStorage` by default — pluggable via `UpToDateStorage` option

## Trade-offs

- **Fast-loop detector vs per-response stale check**: Electric compares response handles against an expired cache. We can't do this (no handles), so we use time-window pattern detection (5+ same-offset requests in 500ms). Less precise per-response but avoids false positives on valid at-tail reads.
- **Pluggable UpToDateStorage vs hardcoded localStorage**: More flexible than Electric's singleton but lacks Electric's 60s write throttle for localStorage (noted for follow-up).
- **ReadableStream pull-based loop vs recursive `#start()`**: Different architecture but equivalent behavior. DS's pull() model is more natural for the ReadableStream API.

## Verification

```bash
# Run client tests (498 pass)
pnpm vitest run --project client

# Run conformance tests

# TypeScript compilation (0 errors in src/)
cd packages/client && pnpm typecheck
```

### New files
| File | Description |
|------|-------------|
| `src/pause-lock.ts` | Set-based multi-reason pause coordination |
| `src/wake-detection.ts` | System sleep/wake detection for Node/Bun |
| `src/fast-loop-detection.ts` | Sliding-window CDN stale response detector |
| `src/up-to-date-tracker.ts` | Pluggable storage for replay mode dedup |
| `SPEC.md` | Formal state machine specification with truth table |
| `test/support/state-machine-dsl.ts` | 6-layer testing DSL with fuzz support |
| `test/support/state-transition-table.ts` | Compiler-enforced 63-cell truth table |

### Modified files
| File | Changes |
|------|---------|
| `src/stream-response-state.ts` | Full rewrite — 7-state hierarchy |
| `src/response.ts` | PauseLock, wake detection, fast-loop, ErrorState recovery, replay, prefetch |
| `src/stream-api.ts` | Split fetch clients, wire onError/replay/prefetch, thread options |
| `src/fetch.ts` | Backoff defaults (1s/2x/32s), header validation, chunk prefetch middleware |
| `src/error.ts` | Add `MissingHeadersError` |
| `src/types.ts` | Add `fastLoopOptions`, `prefetchOptions`, `upToDateStorage` |
| `src/constants.ts` | Add `CACHE_BUSTER_QUERY_PARAM` |

---

Closes #265


samwillis (changes requested):
Thanks for the hardening work here. I found a few issues that look worth fixing before merge.

1. High: SSE closed controls can be dropped or reported as not up-to-date.

`packages/server/src/server.ts` now emits final SSE controls with `streamClosed: true` and intentionally omits `upToDate`, which the protocol allows. But `packages/client/src/response.ts` only emits a control-only synthetic response when `event.upToDate` is truthy, and data+control responses pass `controlEvent.upToDate ?? false`. A valid final close event can therefore either be swallowed entirely or delivered with `streamClosed: true, upToDate: false`.

Relevant code:
- `packages/client/src/response.ts`, in `#processSSEEvents()`: control-only `streamClosed: true` without `upToDate` returns `continue` instead of yielding the terminal response.
- `packages/client/src/response.ts`, in `#processSSEDataEvent()`: the synthetic response uses `controlEvent.upToDate ?? false` even though `streamClosed` implies up-to-date per protocol.

2. Medium: `upToDateStorage` is not updated from SSE controls.

Replay dedup records cursors only in `#updateStateFromResponse()`. SSE controls update `#syncState` but never call `recordUpToDate`, so an SSE live session can advance far past the stored cursor and later replay suppression will use stale data.

3. Medium: pause/wake abort handling can mask real errors.

The catch block in `packages/client/src/response.ts` returns solely because the request controller was aborted with `PAUSE_STREAM` or `SYSTEM_WAKE`, without checking whether the caught error was actually an abort. The skipped test in `packages/client/test/stream-api.test.ts` documents this, and it means a coincident 500/network/protocol error can bypass `onError` and disappear.

I did not run the test suite; this was a review-only pass against `origin/main...HEAD`.

KyleAMathews:
### Code review

Found 3 issues:

1. Closed SSE controls can still be dropped or emitted as not up-to-date. `#updateStateFromSSEControl()` maps `streamClosed: true` to `upToDate: true` internally, but the control-only path only emits when `event.upToDate` is truthy, and the data+control path passes `controlEvent.upToDate ?? false`, so a valid closed SSE control without an explicit `upToDate` can be swallowed or surfaced with `Stream-Closed: true` but no `Stream-Up-To-Date`.

https://github.com/durable-streams/durable-streams/blob/456e97476207f57982ed022584e073d08aa57b27/packages/client/src/response.ts#L685-L703

https://github.com/durable-streams/durable-streams/blob/456e97476207f57982ed022584e073d08aa57b27/packages/client/src/response.ts#L767-L779

2. SSE up-to-date transitions do not update replay storage. Long-poll responses record `recordUpToDate()` in `#updateStateFromResponse()`, but SSE controls update `#syncState` in `#updateStateFromSSEControl()` without recording the cursor, so an SSE session can advance the stored replay cursor in memory only never, leaving later replay suppression stale.

https://github.com/durable-streams/durable-streams/blob/456e97476207f57982ed022584e073d08aa57b27/packages/client/src/response.ts#L457-L481

https://github.com/durable-streams/durable-streams/blob/456e97476207f57982ed022584e073d08aa57b27/packages/client/src/response.ts#L490-L508

3. Consumed prefetched responses are not abortable through body consumption. The prefetch queue now chains the consuming request signal, but removes that listener as soon as the prefetched fetch promise settles. In the current middleware chain, `createFetchWithConsumedBody()` reads `res.arrayBuffer()` after `createFetchWithChunkBuffer()` returns, so a pause/cancel after headers but during body consumption no longer aborts the underlying prefetched response body.

https://github.com/durable-streams/durable-streams/blob/456e97476207f57982ed022584e073d08aa57b27/packages/client/src/fetch.ts#L262-L290

https://github.com/durable-streams/durable-streams/blob/456e97476207f57982ed022584e073d08aa57b27/packages/client/src/fetch.ts#L555-L568

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

KyleAMathews:
Thanks Sam — this was a really useful review.

I verified these with targeted regression tests against the commit you reviewed (`456e9747`) and then re-ran them against the latest PR commit (`e5b6880c`).

Results:

1. **Closed SSE controls without explicit `upToDate`**
   - I added tests for both control-only closed SSE events and data-bearing SSE events followed by `streamClosed: true`.
   - Those tests passed against `456e9747`, so I couldn’t reproduce a user-visible failure here.
   - That said, I still made the synthetic SSE response construction treat `streamClosed` as up-to-date explicitly, which is clearer and safer.

2. **SSE up-to-date transitions not updating replay storage**
   - Confirmed. The regression test failed on `456e9747`: replay storage kept the old cursor (`initial`) instead of recording the SSE up-to-date cursor.
   - Fixed in `e5b6880c` by recording up-to-date cursors from `#updateStateFromSSEControl()` as well as from long-poll response metadata.
   - Regression test now passes.

3. **Consumed prefetched responses not abortable during body consumption**
   - Confirmed. The regression test failed on `456e9747` by timing out because the consumer abort no longer propagated once the prefetched fetch promise had settled.
   - Fixed in `e5b6880c` by wrapping prefetched response bodies with an abort-aware `TransformStream`, keeping the consuming signal active through body consumption.
   - Regression test now passes.

Added verification coverage in:

`packages/client/test/review-293-verification.test.ts`

Latest result:

```text
✓ emits a control-only SSE streamClosed event as up-to-date
✓ marks data-bearing SSE streamClosed controls as up-to-date
✓ records SSE up-to-date cursors in replay storage
✓ aborts a consumed prefetched response while its body is being consumed
```

So: #2 and #3 were real bugs and are fixed; #1 wasn’t reproducible as a failing behavior, but I applied the defensive/clarifying change anyway. Thanks again — especially #2 and #3 were subtle and high-signal.

PR #317, opened 2026-04-02
[Draft]: Stream agent

KyleAMathews:


kevin-dp:
**TODO: Extract tool results from Codex `item/completed` payloads in the normalizer**

The Codex normalizer (`packages/coding-agents/src/normalize/codex.ts`) currently extracts tool **calls** from `item/completed` notifications but discards the tool **results** that are present in the same payload. This matters because the `NormalizedEvent` type system already has a `ToolResultEvent` type, and the Claude normalizer already emits them — but the Codex normalizer never does. The tool call result may be useful for a variety of purposes.

The raw Codex item types already contain the result data (see `packages/coding-agents/src/protocol/codex.ts`):

- `CodexCommandExecutionItem` has `aggregatedOutput` (stdout/stderr) and `exitCode`
- `CodexMcpToolCallItem` has `result` and `error`
- `CodexDynamicToolCallItem` has `contentItems` and `success`
- `CodexFileChangeItem` has `status`

**What to do:**

In `normalizeItem()` in `packages/coding-agents/src/normalize/codex.ts`, for each tool call item type (`commandExecution`, `fileChange`, `mcpToolCall`, `dynamicToolCall`), emit a paired `ToolResultEvent` alongside the existing `ToolCallEvent`. This means `normalizeItem` (and by extension `normalizeCodex`) should be able to return multiple `NormalizedEvent`s — currently they return a single `NormalizedEvent | null`.

The `ToolResultEvent` shape is defined in `packages/coding-agents/src/normalize/types.ts`:

```ts
interface ToolResultEvent {
  type: "tool_result"
  toolCallId: string
  output: string
  isError?: boolean
}
```

Map the Codex fields to this:

- `commandExecution`: `output` = `item.aggregatedOutput`, `isError` = `exitCode !== 0`
- `mcpToolCall`: `output` = `JSON.stringify(item.result)` or `item.error`, `isError` = `!!item.error`
- `dynamicToolCall`: `output` = `JSON.stringify(item.contentItems)`, `isError` = `item.success === false`
- `fileChange`: `output` = `item.status` or similar, `isError` = `status` indicates failure

Update any downstream consumers that call `normalizeCodex()` and expect a single event to handle an array instead.

PR #324, opened 2026-04-09
fix(tests): use eventual assertions for cascade GC

msfstef:
## Summary

- Replace synchronous 404 assertions in cascade GC conformance tests with a `waitForStatus` polling helper

## Why

The protocol spec (PROTOCOL.md §4.2) says cascade garbage collection happens when the last fork is deleted, but uses no RFC 2119 timing keywords — it says data "is cleaned up", not that it "MUST be cleaned up before the DELETE response returns." This makes the timing of cascade GC an implementation detail, not a protocol guarantee.

The two cascade GC tests (`should cascade GC when last fork is deleted` and `should cascade GC through three levels`) were asserting 404 immediately after the DELETE, which is stricter than the protocol requires. A server that performs cascade GC asynchronously (e.g. via a background sweep) would be spec-compliant but fail these tests.

Note: only cascade GC assertions were changed. Soft-delete transitions (410) remain synchronous because the protocol uses MUST for those. Direct deletes (the deleted stream itself returning 404) also remain synchronous.

## Approach

Added a `waitForStatus` utility using `vi.waitFor` with a 5s timeout and 200ms polling interval. No test-level timeout changes — the timeout lives only on the polling helper.

## Test plan

- [x] All 992 conformance tests pass (3 skipped)

### DIFF
--- a/.changeset/eventual-cascade-gc.md
+++ b/.changeset/eventual-cascade-gc.md
@@ -1,0 +1,5 @@
+---
+"@durable-streams/server-conformance-tests": patch
+---
+
+fix: use polling assertions for cascade GC tests instead of synchronous checks to match protocol spec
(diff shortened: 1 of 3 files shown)


balegas (approved):
YOu can improve the working on the protocol too

PR #332, opened 2026-04-13
refactor: replace fastq with zero-dependency AsyncQueue

samwillis:
Replaces the TypeScript client's `fastq` dependency with an internal `AsyncQueue`. This fixes producer flush and automatic epoch-claim paths that call `fastq.drained()`, whose use of `process.nextTick()` fails in browsers without a `process` polyfill.

### Review guidance

`AsyncQueue` uses Promise completion handlers and implements the queue methods used by stream batching and idempotent producers. It includes `running()`, which current main needs to count pending batches and enforce producer backpressure. A head index with periodic compaction keeps dequeue amortized O(1) and releases consumed task references.

The key properties are:

- Tasks start in insertion order, with concurrency bounded by the configured limit.
- `length()` counts waiting tasks; `running()` counts in-flight tasks. Drain waits for both to reach zero, including after worker rejection.
- Invalid concurrency values such as `NaN` and `0.5` still throw at construction.
- Queue push and drain work without the `process` global.

The internal queue covers the methods these callers need, keeping runtime dependencies smaller at the cost of maintaining the scheduling code here. Workers remain Promise-returning, as both client workers already are. Public APIs and protocol behavior stay the same.

### Verification

- Client suite: **333 passed, 3 skipped** including the idle-drain contract and producer auto-claim regression tests.
- TypeScript client conformance: **259 passed, 10 skipped** on the final code.
- Client build and ESLint for the changed source and tests passed.

```sh
pnpm exec vitest run --project client --maxWorkers=2
pnpm --filter @durable-streams/client-conformance-tests test:run
pnpm --filter @durable-streams/client build
pnpm exec eslint packages/client/src/async-queue.ts packages/client/src/idempotent-producer.ts packages/client/src/stream.ts packages/client/test/async-queue.test.ts packages/client/test/idempotent-producer.test.ts
```

Focused unit tests cover ordering, concurrency counts, failure draining, multiple drain waiters, a large backlog and reuse, operation without `process`, and invalid producer concurrency. These exercise internal scheduling and runtime globals that the JSON conformance adapter cannot control. The producer regression removes `process` in Node and holds two responses independently, checking sequence numbers and that `flush()` stays pending until the deferred batch completes. It catches a mutation that stops flush after one drain. A queue test also pins immediate idle-drain behavior when another task is pushed synchronously afterward. These are not tests in a browser engine.

The TypeScript benchmark reports 7 passed and 1 failed even though its job succeeds under `continue-on-error`. The failing “SSE First Event Latency” scenario collects no samples on both this PR and exact base `a172acc3`; it predates the queue replacement. [PR benchmark log](https://github.com/durable-streams/durable-streams/actions/runs/34497758324/job/102940945615), [base benchmark log](https://github.com/durable-streams/durable-streams/actions/runs/29538081794/job/87754206292).

### Files changed

- `packages/client/src/async-queue.ts`: internal queue implementation.
- `packages/client/src/{idempotent-producer,stream}.ts`: replace the queue dependency.
- `packages/client/test/{async-queue,idempotent-producer}.test.ts`: scheduling and validation regressions.
- `packages/client/package.json` and `pnpm-lock.yaml`: remove the client's direct `fastq` dependency.
- `.changeset/`: patch release note for `@durable-streams/client`.


KyleAMathews (approved):
it'd be good to spit out a bunch of tests around how we use it to ensure it doesn't somehow break in the future or have an odd edge case now.

PR #335, opened 2026-04-14
fix file-backed stream tail races on concurrent appends

samwillis:
## Summary
This fixes a race in the file-backed reference server where two append requests could hit the same stream at the same time without producer headers.

In that case, both requests could read the same pre-append `currentOffset`, both write their bytes to disk successfully, and then race when writing LMDB metadata. The loser could overwrite the newer tail with a stale `currentOffset`, leaving metadata behind the real on-disk stream tail. When that happened, live readers and long-polls would compare against the stale metadata, conclude they were already up to date, and miss the newest bytes until restart or recovery reconciled the file tail.

This PR:
- serializes file-backed appends per stream, even for requests that do not use producer headers
- adds a regression test that forces two untagged appends to overlap and verifies the stored tail still advances from `...0003` to `...0006`
- adds a changeset for the `@durable-streams/server` patch release
- preserves the current `main` branch expiry and fork-lifetime behavior while applying the concurrency fix

## Test plan
- [x] Add regression test in `packages/server/test/file-backed.test.ts`
- [x] Run the new test on `main` without the fix and confirm it fails (`currentOffset` stays at `...0003` instead of `...0006`)
- [x] Re-run `pnpm exec vitest run packages/server/test/file-backed.test.ts` with the fix and confirm it passes
- [x] Run `pnpm build` in `packages/server`

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.changeset/fix-file-store-append-race.md
+++ b/.changeset/fix-file-store-append-race.md
@@ -1,0 +1,7 @@
+---
+"@durable-streams/server": patch
+---
+
+fix(server): serialize file-backed appends per stream
+
+File-backed streams could lose track of the latest logical tail when two append requests hit the same stream at the same time without producer headers. The bytes were written to disk, but stale metadata could leave `currentOffset` behind the real file tail, which caused live readers to treat newer updates as already caught up until a restart or recovery reconciled the offsets.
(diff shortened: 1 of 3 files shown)


KyleAMathews (approved):
Can we reproduce the problem in a conformance test? That's much more powerful if so as then it'll prevent the problem in any server implementation. I want us to always default to conformance tests if possible.

PR #344, opened 2026-05-03
fix(client): invoke onError handler during SSE reconnection

HrushiYadav:
Fixes #320.

`onError` was only called during the initial connection attempt. When an established SSE connection dropped and `#trySSEReconnect` ran internally, auth errors (e.g. 401 from an expired token) propagated as uncaught errors with no recovery path, making token refresh during reconnection impossible.

Wrapped the `#startSSE` call inside `#trySSEReconnect` with a retry loop that invokes `onError` on failure and merges any returned headers/params before retrying, matching the existing retry behavior in the initial `stream()` call. Added 3 regression tests covering the 401 + token refresh case, `onError` returning void, and no `onError` handler. All 320 existing tests pass.

### DIFF
--- a/packages/client/src/response.ts
+++ b/packages/client/src/response.ts
@@ -21,11 +21,14 @@ import type { SSEControlEvent, SSEEvent } from "./sse"
 import type { StreamResponseState } from "./stream-response-state"
 import type {
   ByteChunk,
+  HeadersRecord,
   StreamResponse as IStreamResponse,
   JsonBatch,
   LiveMode,
   Offset,
+  ParamsRecord,
   SSEResilienceOptions,
+  StreamErrorHandler,
   TextChunk,
 } from "./types"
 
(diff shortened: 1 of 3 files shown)


HrushiYadav:
Friendly ping -- this PR has been open for ~58 days. The fix ensures the onError handler is invoked during SSE reconnection. Let me know if any changes are needed.

HrushiYadav:
@KyleAMathews @msfstef would either of you have a moment to review? It is a small client fix to invoke the onError handler during SSE reconnection. It is clean and mergeable, and I am happy to add a regression test for the reconnect path if that would help.

HrushiYadav:
Friendly ping — this PR has been open for about 2 months. The fix invokes the onError handler during SSE reconnection. It is clean and mergeable. Happy to add a regression test if that would help move it forward. @KyleAMathews @msfstef

HrushiYadav:
Rebased on latest main. @maintainers — would appreciate a review when you get a chance. Happy to address any feedback.

PR #346, opened 2026-05-04
docs: introduce layered consumer spec into the protocol

icehaunter:
Introduces webhook subscriptions for push-based event delivery, backed by a new layered consumer architecture (L1 named consumers, L2 pull-wake), plus the protocol additions, conformance suite, docs, and server-side perf work that came along the way.

## Protocol (PROTOCOL.md)
- §6 Named Consumers — mechanism-independent consumer identity with epoch fencing, offset tracking, and lease-based liveness (the L1 layer that wake mechanisms are built on).
- §7 Wake-Up Notifications — push-based wake mechanisms over L1, including webhook subscriptions and pull-wake.

New Webhook-Signature header (HMAC-SHA256) registered in §13.
Implements the layered consumer spec

### DIFF
--- a/docs/layered-consumer-spec.md
+++ b/docs/layered-consumer-spec.md
@@ -93,10 +93,10 @@ REGISTERED ───────────────────────
 Pending work exists when any subscribed stream has unprocessed events:
 
 ```
-pending_work = any(tail[path] > acked[path] for path in subscribed_streams)
+pending_work = any(tail[path] >lex acked[path] for path in subscribed_streams)
 ```
 
-Where `acked[path]` is the last acknowledged offset (inclusive — the event at this offset was processed) and `tail[path]` is the current stream end. An acked offset of `-1` means no events have been processed yet. Offset comparison uses the fixed-width lexicographic format defined in the main protocol (see PROTOCOL.md § Offsets).
+Where `>lex` denotes lexicographic comparison (see PROTOCOL.md § 8 — offsets are opaque but lexicographically sortable), `acked[path]` is the last acknowledged offset (inclusive — the event at this offset was processed), and `tail[path]` is the current stream end. An acked offset of `-1` means no events have been processed yet.
 
 ### L1 HTTP API
 
@@ -109,7 +109,6 @@ Content-Type: application/json
 {
   "consumer_id": "my-agent:task-123",
   "streams": ["/agents/task-123"],
-  "namespace": "/agents/*",
   "lease_ttl_ms": 45000
 }
 ```
@@ -118,7 +117,6 @@ Content-Type: application/json
 | -------------- | -------- | ------------------------------------------------------------------------------------------------------------------------------------ |
 | `consumer_id`  | Yes      | Stable, client-provided identifier; must be unique. Must not start with reserved prefix `__wh__:` (used by L2/A synthetic consumers) |
 | `streams`      | Yes      | One or more stream paths to track                                                                                                    |
-| `namespace`    | No       | Glob pattern for informational grouping (e.g., `/agents/*`)                                                                          |
 | `lease_ttl_ms` | No       | Lease duration in milliseconds; server default if omitted                                                                            |
 
 **Response:**
@@ -135,7 +133,6 @@ Content-Type: application/json
   "state": "REGISTERED",
   "epoch": 0,
   "streams": [{ "path": "/agents/task-123", "offset": "-1" }],
-  "namespace": "/agents/*",
   "lease_ttl_ms": 45000
 }
 ```
@@ -171,6 +168,8 @@ POST /consumers/{id}/acquire
   "streams": [{ "path": "/agents/task-123", "offset": "1002" }]
 }
 ```
+
+The `offset` in each stream entry is the last-acknowledged offset (i.e., the consumer's current cursor position). The consumer should resume reading from the next offset after this position.
 
 Acquiring the epoch:
 
@@ -226,7 +225,7 @@ Content-Type: application/json
 
 Both cursor-advancing acks and empty acks reset the lease timer.
 
-**Offset semantics:** Offsets are "last processed inclusive." Offset `"1005"` means events through 1005 have been processed; the next read starts from `"1006"`. Offset `"-1"` means nothing has been processed.
+**Offset semantics:** Offsets are "last processed inclusive." Acking offset `"1005"` means events through that offset have been processed; the consumer resumes reading from the offset returned in `Stream-Next-Offset` of the response that delivered offset `"1005"`. Offset `"-1"` means nothing has been processed.
 
 **Error responses:**
 
@@ -318,7 +317,7 @@ Identity: `__wh__:{subscription_id}:{url_encoded_stream_path}`
 
 The `__wh__:` prefix is a reserved namespace that prevents collision with user-created L1 consumers. The L1 registration endpoint rejects consumer IDs starting with this prefix.
 
-Because a subscription uses a glob pattern (e.g., `/agents/*`), a single subscription can match many streams — each matched stream gets its own consumer instance with independent state, offsets, and lifecycle. The consumer ID encodes the prefix, subscription, and specific stream it tracks. The stream path is URL-encoded to avoid parsing ambiguity. Multiple subscriptions matching the same stream create independent consumers.
+Because a subscription uses a glob pattern (e.g., `/agents/*`), a single subscription can match many streams — each matched stream gets its own consumer instance with independent state, offsets, and lifecycle. This one-consumer-per-matched-stream design exists because each stream represents an independent unit of work (e.g., one agent task) that should wake, process, and idle independently. The consumer can later subscribe to additional streams (e.g., shared filesystem, tool outputs) via the callback API, but it starts with the single stream that triggered its creation. The consumer ID encodes the prefix, subscription, and specific stream it tracks. The stream path is URL-encoded to avoid parsing ambiguity. Multiple subscriptions matching the same stream create independent consumers.
 
 **Implementation note:** When extracting the consumer ID from callback URLs (e.g., `/callback/{consumer_id}`), implementations MUST use the raw percent-encoded path, not a decoded version. HTTP frameworks often decode `%2F` → `/` in URL paths automatically, which would break consumer ID lookups.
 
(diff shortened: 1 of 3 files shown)


docs/layered-consumer-spec.md
```diff
@@ -0,0 +1,827 @@
...
+
+{ "offsets": [] }
+
+→ 200 OK
+{ "ok": true, "token": "eyJ..." }
+```
+
+Both cursor-advancing acks and empty acks reset the lease timer.
+
+**Offset semantics:** Offsets are "last processed inclusive." Offset `"1005"` means events through 1005 have been processed; the next read starts from `"1006"`. Offset `"-1"` means nothing has been processed.
```
icehaunter: This could use better language, as offsets are opaque and starting read from 1005 already would exclude 1005 without knowing what next offset is. Currently it reads as if offsets are incrementable.

docs/layered-consumer-spec.md
```diff
@@ -0,0 +1,827 @@
...
+  "namespace": "/agents/*",
+  "lease_ttl_ms": 45000
+}
+```
+
+| Field          | Required | Description                                                                                                                          |
+| -------------- | -------- | ------------------------------------------------------------------------------------------------------------------------------------ |
+| `consumer_id`  | Yes      | Stable, client-provided identifier; must be unique. Must not start with reserved prefix `__wh__:` (used by L2/A synthetic consumers) |
+| `streams`      | Yes      | One or more stream paths to track                                                                                                    |
+| `namespace`    | No       | Glob pattern for informational grouping (e.g., `/agents/*`)                                                                          |
```
icehaunter: Intention of this field is not clear to me. Patterns in the webhooks are already globs, what is this for?

docs/layered-consumer-spec.md
```diff
@@ -0,0 +1,827 @@
...
+| ------------ | ------------------------------------------------- |
+| `REGISTERED` | Consumer exists; no active reader holds the epoch |
+| `READING`    | Epoch acquired; consumer is actively reading      |
+
+#### Pending Work
+
+Pending work exists when any subscribed stream has unprocessed events:
+
+```
+pending_work = any(tail[path] > acked[path] for path in subscribed_streams)
```
icehaunter: This implies offsets to be comparable, which they are not if opaque

docs/layered-consumer-spec.md
```diff
@@ -0,0 +1,827 @@
...
+
+```http
+POST /consumers/{id}/acquire
+→ 200 OK
+
+{
+  "consumer_id": "my-agent:task-123",
+  "epoch": 3,
+  "token": "eyJ...",
+  "streams": [{ "path": "/agents/task-123", "offset": "1002" }]
```
icehaunter, line 168: Is this offset the latest? Or last acked for this consumer?

docs/layered-consumer-spec.md
```diff
@@ -0,0 +1,827 @@
...
+
+Each consumer tracks its own offsets across one or more streams and cycles through states independently. The underlying L1 consumer record is created at the same time (see L2/A Implementation Notes).
+
+Identity: `__wh__:{subscription_id}:{url_encoded_stream_path}`
+
+The `__wh__:` prefix is a reserved namespace that prevents collision with user-created L1 consumers. The L1 registration endpoint rejects consumer IDs starting with this prefix.
+
+Because a subscription uses a glob pattern (e.g., `/agents/*`), a single subscription can match many streams — each matched stream gets its own consumer instance with independent state, offsets, and lifecycle. The consumer ID encodes the prefix, subscription, and specific stream it tracks. The stream path is URL-encoded to avoid parsing ambiguity. Multiple subscriptions matching the same stream create independent consumers.
```
icehaunter: What's the benefit here of this single-consumer-per-stream given the consumer definition can take N streams? Or is a consumer always holding a lease over the entire provided stream list? If that's the case, it wasn't clear from the text, we should update it

docs/layered-consumer-spec.md
```diff
@@ -0,0 +1,827 @@
...
+                    │ done+¬pending │  LIVE   │───┘    │     │
+                    │ OR lease exp. │(L1:READ)│        │     │
+                    └───────────────┴─────────┘        │     │
+                    │                                   │     │
+                    │  done + pending_work               │     │
+                    └───────────────────────────────────┘     │
+                                                              │
+                    10s timeout, no 2xx, no callback ──────────┘
+                    (retry webhook delivery)
+```
```
icehaunter, line 378: This ascii makes no sense, which is a shame because it would have been helpful

docs/layered-consumer-spec.md
```diff
@@ -0,0 +1,827 @@
...
+
+#### State Transitions
+
+| From   | To      | Trigger                                                          | Side Effects                            |
+| ------ | ------- | ---------------------------------------------------------------- | --------------------------------------- |
+| IDLE   | WAKING  | `pending_work` becomes true                                      | L1 acquire(), new wake_id, webhook POST |
```
icehaunter: It's not clear to me if `pending_work` here is a real state variable that we need to actually implement as a reactive property, or if this is a document-specific shorthand to "work is available". Because I'm not sure the "consumer" is specced to expose this property in this doc

docs/layered-consumer-spec.md
```diff
@@ -0,0 +1,827 @@
...
+| `triggered_by`   | string[] | Streams that have pending events (informational) |
+| `callback`       | string   | URL for acknowledgments and subscription changes |
+| `token`          | string   | L1 bearer token for the first callback           |
+
+### Synchronous Done
+
+The webhook response may include `{ "done": true }` to immediately return to IDLE without using the callback API. When the server receives this response:
+
+1. The wake is considered claimed
+2. All streams are auto-acked to their current tail offset
```
icehaunter: If new events arrive while the handler is processing, or between its final read and the server’s auto-ack, those events can be acknowledged without being processed.

docs/layered-consumer-spec.md
```diff
@@ -0,0 +1,827 @@
...
+  }
+}
+```
+
+### Worker Flow
+
+1. Worker long-polls the wake stream for new events
+2. On receiving a `wake` event, worker calls `POST /consumers/{id}/acquire`
+3. If acquire succeeds, worker reads events, acks progress, and releases when done
+4. If acquire fails (`EPOCH_HELD` or `INTERNAL_ERROR`), another worker claimed it — skip
```
icehaunter: How are we distinguishing workers in the competitive case vs self-supersede? Worker identity in self-supersede is likely unstable (as in after the crash as mentioned above), so "acquire fails" is underdefined here

icehaunter (commented):
@KyleAMathews

docs/layered-consumer-spec.md
KyleAMathews: 👍 yeah sounds good let's improve

docs/layered-consumer-spec.md
KyleAMathews: removed

docs/layered-consumer-spec.md
KyleAMathews: there can be multiple readers but only one consumer at a time can "claim" the stream.

docs/layered-consumer-spec.md
KyleAMathews: pushed update

docs/layered-consumer-spec.md
KyleAMathews: I'm pretty sure not? If the handler is running, any newly arriving events would be processed in the next run through the handler.

docs/layered-consumer-spec.md
msfstef: last-acked offset for the consumer it seems

docs/layered-consumer-spec.md
```diff
@@ -0,0 +1,828 @@
...
+  "streams": ["/agents/task-123"],
+  "lease_ttl_ms": 45000
+}
+```
+
+| Field          | Required | Description                                                                                                                          |
+| -------------- | -------- | ------------------------------------------------------------------------------------------------------------------------------------ |
+| `consumer_id`  | Yes      | Stable, client-provided identifier; must be unique. Must not start with reserved prefix `__wh__:` (used by L2/A synthetic consumers) |
+| `streams`      | Yes      | One or more stream paths to track                                                                                                    |
+| `lease_ttl_ms` | No       | Lease duration in milliseconds; server default if omitted                                                                            |
```
msfstef, line 120: might be nice to add a TTL/Expiry mechanic here so that consumers are not created and never cleaned up (or require client side book keeping to ensure clean ups)

docs/layered-consumer-spec.md
```diff
@@ -0,0 +1,828 @@
...
+Content-Type: application/json
+
+{
+  "offsets": [
+    { "path": "/agents/task-123", "offset": "1005" }
+  ]
+}
+
+→ 200 OK
+{ "ok": true, "token": "eyJ..." }
```
msfstef, line 210: should we be using signed URLs here as well instead of bearer tokens?

docs/layered-consumer-spec.md
```diff
@@ -0,0 +1,828 @@
...
+
+**S6 — Stale Epoch Rejection**: Ack requests with a token epoch that does not match the current consumer epoch are rejected with `STALE_EPOCH`. This fences zombie consumers — a session from a previous epoch cannot interfere with the current one.
+
+**S9 — Critical Callback Atomicity**: If a critical L2 callback fails during acquire, the acquire is rolled back (epoch released) and the caller receives `INTERNAL_ERROR`. L1 and L2 state remain consistent — no acquire completes without its critical L2 side effects succeeding.
+
+**S10 — Reserved Prefix**: Consumer IDs starting with `__wh__:` are reserved for L2/A synthetic consumers. The L1 registration endpoint rejects consumer IDs with this prefix to prevent namespace collision.
+
+### L1 Liveness Properties
+
+**LP1 — Lease Expiry Causes Epoch Release**: When the lease timer fires while the consumer is READING, the epoch is released (READING → REGISTERED). Any registered `onLeaseExpired` callbacks (e.g., the L2 webhook layer) are invoked.
```
msfstef, line 273: Would the lease expiry release fail if `onLeaseExpired` fails, similar to how it would fail to acquire if the L2 callbacks failed?

docs/layered-consumer-spec.md
```diff
@@ -0,0 +1,828 @@
...
+
+- `subscription_id` — Client-provided identifier
+- `pattern` — Glob pattern matching stream paths
+- `webhook` — URL to POST notifications to
+- `webhook_secret` — Server-generated HMAC secret (returned on creation only)
+- `description` — Optional human-readable label
+
+#### Consumer Instance
+
+Created **lazily on first append** to a stream matching a subscription's pattern. Consumer instances are not created when a stream is first created or when a subscription is registered against existing streams — only when actual events arrive. This avoids materializing consumer records for streams that may never receive data (e.g., 100k agent streams where only a fraction are active).
```
msfstef, line 312: This means that on every stream, on every append, there needs to be some kind of matching and check of whether consumers need to be instantiated. Unclear on how this would scale and what systems we would use to perform this matching, a database is probably needed.

docs/layered-consumer-spec.md
```diff
@@ -0,0 +1,828 @@
...
+
+Identity: `__wh__:{subscription_id}:{url_encoded_stream_path}`
+
+The `__wh__:` prefix is a reserved namespace that prevents collision with user-created L1 consumers. The L1 registration endpoint rejects consumer IDs starting with this prefix.
+
+Because a subscription uses a glob pattern (e.g., `/agents/*`), a single subscription can match many streams — each matched stream gets its own consumer instance with independent state, offsets, and lifecycle. This one-consumer-per-matched-stream design exists because each stream represents an independent unit of work (e.g., one agent task) that should wake, process, and idle independently. The consumer can later subscribe to additional streams (e.g., shared filesystem, tool outputs) via the callback API, but it starts with the single stream that triggered its creation. The consumer ID encodes the prefix, subscription, and specific stream it tracks. The stream path is URL-encoded to avoid parsing ambiguity. Multiple subscriptions matching the same stream create independent consumers.
```
msfstef, line 320: it state steh consumer can later subscribe to additional streams - but we don't have tha option, and we should somehow decide if we do add that how it would work with lease acquisition etc, would only the lease holder be able to edi t the consumer? it feels like it might open a can of worms, would avoid it if possible

docs/layered-consumer-spec.md
msfstef: unfortunately the ascii diagram still doesn't make complete sense - missing live -> waking path

docs/layered-consumer-spec.md
```diff
@@ -0,0 +1,828 @@
...
+                └──────────────────────────────────────────┘
+                    (L1 release, auto-ack to tail)
+```
+
+#### State Transitions
+
+| From   | To      | Trigger                                                          | Side Effects                            |
+| ------ | ------- | ---------------------------------------------------------------- | --------------------------------------- |
+| IDLE   | WAKING  | `pending_work` detected (on append to subscribed stream)         | L1 acquire(), new wake_id, webhook POST |
+| WAKING | LIVE    | Webhook responds 2xx, OR callback claims wake_id                 | L1 lease timer running                  |
```
msfstef, line 385: wouldn't the acquire() call succeeding have the lease timer running as a side effect? or is this where acquire() wouldn't complete untl the webhook responds?

msfstef (commented):
first round of comments - will do one more tomorrow

PR #351, opened 2026-05-14
fix(cli): allow slash-delimited stream IDs

KyleAMathews:
## Summary

Rewrites the CLI to support slash-delimited stream IDs (e.g., `org/project/stream`) and adds proper flag parsing, global options, JSON mode, and auth support. Previously, stream IDs containing `/` were rejected by validation or misinterpreted during argument parsing.

## Root cause

The original CLI had no explicit stream ID validation and used a naive positional argument parser. Stream IDs like `account/chat/room-1` would either fail validation or get split across arguments. Additionally, flags like `--json` placed *before* the stream ID (a natural usage pattern) were not handled.

## Approach

**Argument parsing** — Replaced positional-only parsing with a two-phase approach:
1. `parseGlobalOptions` extracts `--url` and `--auth` from anywhere in the arg list, falling back to `STREAM_URL`/`STREAM_AUTH` env vars
2. `getStreamIdAndArgs` allows command-specific flags (`--json`, `--content-type`, `--batch-json`) to appear before or after the stream ID

**Validation** — New `validation.ts` module with `validateStreamId` using the pattern `/^[a-zA-Z0-9_\-.:/]+$/` — explicitly allowing forward slashes for hierarchical IDs.

**URL construction** — `buildStreamUrl(baseUrl, streamId)` simply concatenates, so slashes in the stream ID naturally form the correct URL path.

## Key invariants

- Stream IDs with slashes produce correct URLs: `base/org/project/stream`
- `--url` and `--auth` are stripped before command-specific parsing, so they don't interfere with stream ID detection
- Flags before the stream ID are collected and passed through to command-specific parsers

## Non-goals

- No changes to the server or protocol — slash-delimited IDs are already valid URL paths
- No changes to any client libraries

## Trade-offs

The CLI is substantially rewritten rather than patching the old parser. This was necessary because the old code had no flag parsing infrastructure, no validation, and hardcoded the stream URL format — a surgical fix for just slashes would have been fragile.

## Verification

```bash
pnpm vitest run packages/cli/test/validation.test.ts
pnpm vitest run packages/cli/test/cli.test.ts
pnpm --filter @durable-streams/cli build
```

## Files changed

| File | Description |
|------|-------------|
| `packages/cli/src/index.ts` | Full rewrite: global options, two-phase arg parsing, JSON/batch-json modes, better error messages |
| `packages/cli/src/validation.ts` | New validation module — URL, auth, stream ID (with slash support) |
| `packages/cli/test/cli.test.ts` | Integration tests including slash-delimited IDs and flags-before-stream-ID |
| `packages/cli/test/validation.test.ts` | Unit tests for all validation functions |
| `docs/cli.md` | New CLI documentation page with hierarchical stream ID examples |
| `packages/cli/README.md` | Updated README with new options and examples |

### DIFF
--- a/packages/cli/src/validation.ts
+++ b/packages/cli/src/validation.ts
@@ -105,7 +105,7 @@ export function validateAuth(auth: string): ValidationResult {
 /**
  * Validate a stream ID.
  * Must be 1-256 characters containing only: letters, numbers, underscores,
- * hyphens, dots, and colons (URL-safe characters).
+ * hyphens, dots, colons, and forward slashes (URL path-safe characters).
  */
 export function validateStreamId(streamId: string): ValidationResult {
   if (!streamId || !streamId.trim()) {
@@ -115,12 +115,13 @@ export function validateStreamId(streamId: string): ValidationResult {
     }
   }
 
-  // Stream IDs should be URL-safe
-  const validPattern = /^[a-zA-Z0-9_\-.:]+$/
+  // Stream IDs should be URL path-safe. Slashes are allowed so callers can
+  // address hierarchical stream IDs such as "account/chat/room-1".
+  const validPattern = /^[a-zA-Z0-9_\-.:/]+$/
   if (!validPattern.test(streamId)) {
     return {
       valid: false,
-      error: `Invalid stream ID: "${streamId}"\n  Stream IDs can only contain letters, numbers, underscores, hyphens, dots, and colons`,
+      error: `Invalid stream ID: "${streamId}"\n  Stream IDs can only contain letters, numbers, underscores, hyphens, dots, colons, and slashes`,
     }
   }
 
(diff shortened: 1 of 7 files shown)


packages/cli/src/validation.ts
```diff
@@ -105,7 +105,7 @@ export function validateAuth(auth: string): ValidationResult {
 /**
  * Validate a stream ID.
  * Must be 1-256 characters containing only: letters, numbers, underscores,
- * hyphens, dots, and colons (URL-safe characters).
+ * hyphens, dots, colons, and forward slashes (URL path-safe characters).
```
msfstef, line 108: _cólons_

msfstef (approved):
LGTM - needs changeset for this?

PR #356, opened 2026-05-19
fix(transports): engage client batching by not awaiting each append (#355)

### LINKED ISSUE #355: Per-chunk await on stream.append disables batching and slows streaming in AI SDK / TanStack AI transports
### The pattern that slows it down

The AI SDK transport (`@durable-streams/aisdk-transport@0.0.3`, `src/server.ts`, `writeSourceToStream`) drives the source iterable like this:

```ts
for await (const chunk of source) {
  await stream.append(JSON.stringify(chunk), { contentType });
}
```

The TanStack AI transport — the equivalent chat transport on the TanStack side — uses the same `for await … await stream.append(...)` shape on the server side. The `@durable-streams/client` README examples teach the same pattern (`await stream.append(JSON.stringify({ message: "hello" }))`). It is the documented, idiomatic path on every consumer, and it serializes every chunk into its own HTTP roundtrip.

### Why it slows streaming down

`@durable-streams/client@0.2.3` (`src/stream.ts`) ships with batching on by default:

- The constructor sets up a `fastq` queue with **concurrency 1**.
- `#appendWithBatching` pushes each call into an in-memory `#buffer` and only enqueues a POST when the queue is **idle**.
- The batch worker, upon completing a POST, drains whatever accumulated in `#buffer` during that POST into the next POST.

That batching only kicks in if multiple `append()` calls overlap in time. With the documented `await`-per-chunk usage, the loop is strictly serialized: by the time the next iteration calls `append`, the queue is always idle and `#buffer` is always empty. Every chunk becomes a single-item POST, and the batching machinery is dead code.

Throughput stops being LLM-bound and becomes HTTP-bound — one network roundtrip per token chunk.

### Approximate impact

On either transport (AI SDK or TanStack AI) with a JSON content type, with a typical durable-streams endpoint:

- Each `stream.append` POST roundtrip is on the order of **a few tens of milliseconds**.
- That caps the writer at roughly **20 chunks/sec**.
- Many modern LLMs emit at **40–50+ chunks/sec** when streaming.
- The writer falls behind the model, and there is a long tail of HTTP appends after the LLM has finished. For a moderate-length generation this can easily be **tens of seconds of tail latency**. End-to-end, the UI is still streaming tokens long after the model has finished generating.

### Root cause

The API has two conflicting properties:

1. The batching feature is the headline performance optimization, and it's correctly implemented internally (`fastq` + `#buffer` + idle-queue trigger).
2. The documented usage pattern (`await` each call) makes it impossible to engage. There's no `lingerMs` / debounce on `append()` itself — that timing parameter exists only on `appendStream()` / `writable()` via `IdempotentProducer`.

`appendStream` and `writable` work fine because they pipe a `ReadableStream` over a single POST. But the AI SDK transport and the TanStack AI transport both have an `AsyncIterable<unknown>` where each item needs JSON-per-item semantics (so the client can re-emit discrete chat chunks). That naturally reaches for `append()`, and `append()` is where the slowdown lives.

### Possible upstream fixes (rough order of intrusiveness)

1. **Documentation**: tell users not to `await` individual `append()` calls in tight loops, and to drain via a final `flush()` / `close()`. The current README examples actively teach the slow pattern.
2. **Add `lingerMs` to `append()`** (mirroring `IdempotentProducer`): a small linger (~5–10 ms) before firing the first POST coalesces a burst of awaited callers into a single batch with zero caller changes.
3. **Fix both transports directly**: in the AI SDK transport and the TanStack AI transport's `writeSourceToStream`, stop awaiting each `append()`. Track only the latest pending promise (the internal queue is FIFO + concurrency 1, so awaiting the last drains all earlier ones), and `await` it once in `finally` before `stream.close()`. Capture per-append errors via `.catch` and re-raise after close. This is a localized change in each transport and gives every consumer the win for free.

### TL;DR

The default batching path requires concurrent (non-awaited) callers, but every example, the AI SDK transport, and the TanStack AI transport all `await` every append. Batching never engages, the writer is capped well below typical LLM emission rates, streaming visibly slows down, and real generations leave noticeable tail latency after the model has already finished.

balegas:
Closes #355.

## Summary

The AI SDK and TanStack AI transports wrote every chunk with `await stream.append(...)`, which kept the client's built-in batching queue idle and turned every LLM token into its own HTTP roundtrip. The transport response writers now fire-and-track appends and drain before close, restoring batching for high-frequency streams. User-visible impact: tail latency after the model finishes generating drops from "tens of seconds" (one POST RTT per chunk × hundreds of chunks) to roughly one RTT.

## Root cause

`DurableStream.append()` in `@durable-streams/client` ships with batching enabled: a `fastq` queue with concurrency 1 buffers calls that arrive while a POST is in flight, then flushes them together when the worker drains. The mechanism only engages when callers don't serialize themselves. The transports' source loops looked like:

```ts
for await (const chunk of source) {
  await stream.append(JSON.stringify(chunk), { contentType })
}
```

Awaiting each call means the queue is always idle on entry — `#appendWithBatching` fires a single-item POST every iteration and the buffer/worker code path is never exercised. Throughput collapses from "LLM-bound" to "HTTP-bound at ~1/RTT per chunk."

## Approach

Each transport's response writer now keeps a `lastAppend: Promise<void>` cursor. The loop fires `append()` without awaiting, attaches a `.catch` that latches the first error into a closure variable (and prevents unhandled rejections for fire-and-forget calls), and short-circuits on error. Before `close()`, we await `lastAppend` — the client queue is FIFO with concurrency 1, so awaiting the latest tracked promise drains every earlier one too. Append errors take precedence over source errors when both occur.

```ts
let appendError: unknown
let lastAppend: Promise<void> = Promise.resolve()
for await (const chunk of source) {
  if (appendError !== undefined) break
  lastAppend = stream.append(JSON.stringify(chunk), { contentType })
    .catch((err) => { if (appendError === undefined) appendError = err })
}
await lastAppend  // drains FIFO queue
await stream.close()
if (appendError !== undefined) throw appendError
```

Same shape applied to `writeSourceToStream`, `pipeSanitizedChunksToStream`, and `appendSanitizedChunksToStream`.

## Key invariants

- **Order**: writes still hit the server in source order. The client queue is FIFO with concurrency 1, so coalescing batches multiple items into a single POST without reordering across POSTs.
- **Durability**: `close()` only runs after every `append()` has resolved — readers don't see a closed stream missing the tail.
- **Error visibility**: any rejected `append()` is surfaced to the caller. The first error wins; subsequent appends short-circuit the source loop.
- **No unhandled rejections**: every `append()` promise has a `.catch` attached at issue time, so fire-and-forget calls don't leak rejection warnings.

## Non-goals

- **No `lingerMs` on `append()`** (option 2 in the issue). The transport-side fix gets the win without a timer-driven debounce; adding `lingerMs` would change latency semantics for callers who deliberately await each append for backpressure. Worth a follow-up issue if anyone wants automatic batching for user code that still serializes.
- **StateDB docs unchanged**. `docs/stream-db.md` shows `await stream.append(...)` per logical mutation, which is the right shape for that consumer — those aren't tight token loops.
- **No conformance test changes**. This is TS-only transport-adapter behaviour; the protocol layer is already covered.

## Trade-offs considered

- **Fix the transports** (this PR) vs. **document around the slow pattern** (the issue's option 1) vs. **add `lingerMs` to `append()`** (option 2). The transports are the high-traffic consumers and a localized change unblocks every downstream user without API surface changes. Docs alone don't help users who never see the warning. `lingerMs` is more general but heavier and changes caller-visible timing. We also updated client JSDoc/README so user code doesn't reintroduce the pattern.

## Verification

```bash
# Full suite — now includes both transport projects
pnpm test:run
# → 1014 passed, 3 skipped, 0 failed (was 1000 before)

# Transport tests alone
pnpm vitest run --project tanstack-transport --project aisdk-transport
# → 14 passed (5 tanstack writeSourceToStream + 1 pipe + 1 appendSanitized + 3 aisdk + 4 pre-existing session)

# Typecheck + build
pnpm --filter @durable-streams/aisdk-transport typecheck
pnpm --filter @durable-streams/tanstack-ai-transport typecheck
pnpm --filter @durable-streams/aisdk-transport build
pnpm --filter @durable-streams/tanstack-ai-transport build
```

The TDD tests verify the load-bearing property by mocking `DurableStream` so every `append()` stays pending until the source iterator finishes, then asserting all chunks were fired before any append resolved. Pre-fix, these tests fail with `expected 5 to be 1`. Post-fix, they pass.

## Files changed

| File | Change |
|---|---|
| `packages/aisdk-transport/src/server.ts` | Fire-and-track-last refactor of `writeSourceToStream` |
| `packages/aisdk-transport/test/server.test.ts` | New — 3 tests covering concurrency, drain-before-close, error propagation |
| `packages/aisdk-transport/package.json`, `tsconfig.json` | Wire up vitest + include `test/**` |
| `packages/tanstack-ai-transport/src/server.ts` | Same refactor for `writeSourceToStream`, `pipeSanitizedChunksToStream`, `appendSanitizedChunksToStream` |
| `packages/tanstack-ai-transport/test/server.test.ts` | New — 5 tests covering the three helpers |
| `packages/client/src/stream.ts` | JSDoc now documents batching engagement contract; recommends `appendStream()` / `writable()` for tight loops |
| `packages/client/README.md` | Performance note added to the `append()` example |
| `vitest.config.ts`, `package.json` | New `aisdk-transport` vitest project; `test:run` now includes both transport projects |
| `.changeset/transport-append-batching.md` | Patch bump for `client`, `aisdk-transport`, `tanstack-ai-transport` |

## Test plan

- [x] `pnpm test:run` passes (1014/1014)
- [x] Transport vitest projects pass (14/14)
- [x] Both transports typecheck clean
- [x] Both transports build clean
- [x] Changeset present and `pnpm changeset status` lists the three packages


samwillis (approved):
LGTM :shipit:

PR #358, opened 2026-05-21
fix(caddy-plugin): auto-truncate corrupt segment tails during recovery

KyleAMathews:
## Summary

`RecoverStore` now truncates trailing corrupt bytes from segment files during startup recovery. Previously, partial writes from crashes left bogus length prefixes that caused `ErrCorruptedSegment` on subsequent reads — even though all valid data before the corruption was perfectly recoverable.

**User-visible impact**: Servers that crash mid-write now self-heal on restart instead of failing with corrupted segment errors.

## Root cause

The segment file format is `[4-byte length][data bytes]` repeated. A crash mid-write can leave a partial length prefix or truncated message body at the tail. `ScanSegment()` already correctly identifies the last valid byte offset by re-reading frames from the start, but `RecoverStore` never acted on a size mismatch — it reconciled the metadata offset but left the corrupt tail bytes on disk. The next `Append` or `Read` would then hit the garbage bytes and return `ErrCorruptedSegment`.

## Approach

1. **Wire `RecoverStore` into production** — Called from `Provision()` before `NewFileStore` so recovery actually runs on server startup (it was previously dead code)
2. **Single file descriptor** — `recoverSegment()` opens the segment once, scans via `ScanSegmentFile()`, stats, truncates, and syncs through the same fd — eliminating the TOCTOU race between stat and truncate
3. **Structured recovery events** — `RecoverStoreWithEvents()` accepts an `onEvent` callback. The Caddy module logs truncations through zap with stream path, segment path, original/recovered sizes, and discarded byte count
4. **Defensive guard** — Uses `>` not `!=` so truncation only removes excess bytes, never zero-extends a smaller-than-expected file

## Key invariants

- `ScanSegment` can never return a `ByteOffset` larger than the file size (it breaks on EOF/short reads)
- Truncation is idempotent — re-running recovery on an already-recovered store is a no-op
- `f.Sync()` after truncate ensures the repair survives a subsequent crash before the OS flushes

## Non-goals

- **Graceful degradation in `ReadMessages`**: The existing reader still returns `ErrCorruptedSegment` on corrupt data rather than returning the valid prefix. That's a separate concern — this PR ensures corruption is cleaned up before the reader ever sees it.
- **Additional corruption patterns in tests**: The two tests cover the core scenario (bogus length prefix) and structured event reporting. `ScanSegment` already handles all corruption variants (partial length, truncated body, etc.) — the truncation logic doesn't care *how much* garbage is at the tail.

## Verification

```bash
cd packages/caddy-plugin && go test ./... -v
```

The `TestRecoverStore_TruncatesCorruptSegmentTail` test creates a valid segment, appends a corrupt length prefix, runs recovery, verifies the file was truncated to the valid size, then confirms the store accepts new appends and reads back all messages. `TestRecoverStoreWithEvents_ReportsTruncation` verifies the structured event callback reports correct sizes.

## Files changed

| File | Change |
|------|--------|
| `store/segment.go` | Extract `ScanSegmentFile()` so recovery can scan via an open fd |
| `store/file_store.go` | `RecoveryEvent` struct, `RecoverStoreWithEvents()`, `recoverSegment()` helper |
| `module.go` | Call `RecoverStoreWithEvents` in `Provision` before `NewFileStore` |
| `store/file_store_test.go` | Two tests: truncation + read-after-recovery, and event reporting |

### DIFF
--- a/packages/caddy-plugin/module.go
+++ b/packages/caddy-plugin/module.go
@@ -66,6 +66,16 @@ func (h *Handler) Provision(ctx caddy.Context) error {
 		h.logger.Info("using in-memory store (no data_dir configured)")
 	} else {
 		// Use file-backed store
+		if err := store.RecoverStoreWithEvents(h.DataDir, func(event store.RecoveryEvent) {
+			h.logger.Warn("truncated corrupt segment tail during recovery",
+				zap.String("stream", event.StreamPath),
+				zap.String("segment", event.SegmentPath),
+				zap.Uint64("original_size", event.OriginalSize),
+				zap.Uint64("recovered_size", event.RecoveredSize),
+				zap.Uint64("discarded_bytes", event.DiscardedBytes))
+		}); err != nil {
+			return fmt.Errorf("failed to recover store: %w", err)
+		}
 		fileStore, err := store.NewFileStore(store.FileStoreConfig{
 			DataDir:        h.DataDir,
 			MaxFileHandles: h.MaxFileHandles,
--- a/packages/caddy-plugin/store/file_store.go
+++ b/packages/caddy-plugin/store/file_store.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"crypto/rand"
 	"encoding/hex"
+	"errors"
 	"fmt"
 	"net/url"
 	"os"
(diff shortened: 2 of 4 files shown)


icehaunter (approved):
Lgtm as far description goes, I don't read fluent go

PR #362, opened 2026-05-25
feat(caddy): add webhook subscriptions

icehaunter:
Adds Caddy plugin webhook subscription support as the top PR in the stack, after the shared protocol and TypeScript server APIs.

Motivation and decisions:
- Keep the Caddy implementation isolated from the shared protocol/server PR so it can be reviewed independently.
- Add webhook subscription storage, stream matching, callback routing, retry handling, acknowledgement routes, and JWKS-backed Ed25519 request signatures.
- Wire webhook configuration through the Caddy module and extend the Caddy conformance setup.
- Include an empty changeset because the Caddy plugin package is internal and excluded from published package versioning.


samwillis (approved):
LGTM

PR #371, opened 2026-05-28
fix(client): serialize auto-claim producer batches

icehaunter:
## Summary

Fixes an idempotent producer race when autoClaim is enabled and multiple batches are appended before the first request has finished claiming its epoch.

Previously, the producer only checked queued work with fastq.length, which does not include the currently running batch. A later batch could therefore be sent while the initial auto-claim request was still in flight. Because sequence numbers were reserved before the claim completed, the later request could become the first request seen for the newly claimed epoch with a non-zero sequence number, causing the server to reject it with New epoch must start with sequence 0.

This change:

- Counts both queued and running fastq work in inFlightCount.
- Defers later auto-claim batches until the first running batch has completed the epoch claim.
- Reserves sequence numbers only when a deferred batch is actually pushed.
- Makes flush wait for batches held behind the initial auto-claim barrier.
- Adds a regression test that verifies later batches are not sent before the first auto-claiming batch completes, and that flush waits for the delayed batch.

This was found while validating Electric Agents against the latest published Durable Streams packages.

### DIFF
--- a/packages/client/src/idempotent-producer.ts
+++ b/packages/client/src/idempotent-producer.ts
@@ -145,6 +145,7 @@ export class IdempotentProducer {
   // Pipelining via fastq
   readonly #queue: queueAsPromised<BatchTask>
   readonly #maxInFlight: number
+  readonly #deferredEnqueues = new Set<Promise<void>>()
   #closed = false
   #closeResult: CloseResult | null = null
   #pendingFinalMessage?: Uint8Array | string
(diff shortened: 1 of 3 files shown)


packages/client/src/idempotent-producer.ts
```diff
@@ -325,8 +326,15 @@ export class IdempotentProducer {
...
     }
 
-    // Wait for queue to drain
-    await this.#queue.drained()
+    // Wait for the queue and any batches deferred behind the auto-claim
+    // barrier. A deferred enqueue can push more queue work after a drain.
+    do {
+      await this.#queue.drained()
+
+      if (this.#deferredEnqueues.size > 0) {
```
kevin-dp: Is this if test really needed?
If the array is empty `Promise.all` will resolve to an empty array.

kevin-dp (approved):

PR #382, opened 2026-06-03
feat(state)!: move TanStack DB surface to @durable-streams/state/db subpath

msfstef:
## Why

#381 correctly made `@tanstack/db` a **peer dependency** of `@durable-streams/state` — it's a singleton (collections/transactions/live queries rely on `instanceof` and module-level state), so two copies in a tree don't interoperate.

But the package's main entry **eagerly re-exported ~30 runtime values** from `@tanstack/db`, so importing *anything* from `@durable-streams/state` forced that peer to be resolvable at module-load time.

That broke publishable dependents that don't use the reactive layer. The clearest case is **`@durable-streams/server`**, which imports only `createStateSchema` (pure schema validation — no `@tanstack/db` at runtime) yet crashed on import:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find package '@tanstack/db'
  imported from .../packages/state/dist/index.js
```

for any consumer that hadn't installed the unmet peer. npm v7+ auto-installs peers, but **pnpm and yarn do not** — so a published `server` was effectively broken for them.

## What

Split the package along the protocol's natural **write/read boundary**:

- **`@durable-streams/state`** (main entry) is now **free of `@tanstack/db`**. It exposes the producer/backend surface: `createStateSchema` + change-event helpers, `MaterializedState`, and the state-event types/guards. Backends like the server depend on it without the peer.
- **`@durable-streams/state/db`** (new subpath) holds the reactive, TanStack DB-backed layer: `createStreamDB`, `getStreamDBCollectionId`, the `StreamDB*`/`Action*` types, and the `@tanstack/db` convenience re-exports. It's a **strict superset** of the main entry, so reactive consumers only change the import path:

```diff
-import { createStateSchema, createStreamDB } from "@durable-streams/state"
+import { createStateSchema, createStreamDB } from "@durable-streams/state/db"
```

### Mechanics
- Extracted the db-free schema/event-construction code into `src/schema.ts`.
- `src/index.ts` re-exports only the db-free surface; `src/db.ts` is the new superset entry; `stream-db.ts` re-exports `createStateSchema` for back-compat (kept the large existing test untouched).
- Added the `./db` export-map entry and a second tsdown build entry.
- Updated the four examples that use `createStreamDB` to import from `/db` (one-line path change each).

## Testing
- New `test/db-free-entry.test.ts` traces each entry's **module graph** and asserts the main entry never imports `@tanstack/db` while `/db` does. It failed before the split and passes now.
- Verified the built `server` imports cleanly with `@tanstack/db` removed from the tree (previously `ERR_MODULE_NOT_FOUND`).
- `state` typechecks; 67/67 state tests pass; `/db` subpath typechecks in the `wikipedia-events`, `test-ui`, and `yjs-demo` examples.

## Dependency notes
- No **publishable** package needs `@tanstack/db` added — none consume the reactive `/db` layer.
- The `/db`-consuming examples are all `private: true`, don't import `@tanstack/db` directly, and the root `pnpm.overrides` already pins `@tanstack/db` to a single monorepo-wide version, so no explicit dep is added there.

## Breaking change
`createStreamDB`, `getStreamDBCollectionId`, the StreamDB/Action types, and the `@tanstack/db` re-exports are no longer exported from `@durable-streams/state` — import them from `@durable-streams/state/db`. The db-free APIs are unchanged on the main entry. Released as a `minor` bump (pre-1.0).

### DIFF
--- a/.changeset/state-db-subpath.md
+++ b/.changeset/state-db-subpath.md
@@ -1,0 +1,21 @@
+---
+"@durable-streams/state": minor
+---
+
+feat(state)!: move the TanStack DB-backed surface to a `@durable-streams/state/db` subpath
+
+The main `@durable-streams/state` entry is now free of `@tanstack/db`. It exposes only the db-free protocol surface — `createStateSchema` and its change-event helpers, `MaterializedState`, and the state-event types/guards — so producers and backends (e.g. `@durable-streams/server`) can depend on it without the `@tanstack/db` peer dependency installed. Previously the entry eagerly re-exported `@tanstack/db`, so importing anything from the package forced that peer to be resolvable, breaking publishable dependents that don't use the reactive layer.
+
+The reactive, TanStack DB-backed layer now lives at `@durable-streams/state/db`:
+
+- `createStreamDB`, `getStreamDBCollectionId`, and the `StreamDB*` / `Action*` types
+- the convenience re-exports of `@tanstack/db` (`createCollection`, `createOptimisticAction`, `eq`, `and`, `count`, …)
+
+The subpath is a strict superset of the main entry (it also re-exports `createStateSchema`), so existing reactive consumers only change the import path:
+
+```diff
+-import { createStateSchema, createStreamDB } from "@durable-streams/state"
++import { createStateSchema, createStreamDB } from "@durable-streams/state/db"
+```
+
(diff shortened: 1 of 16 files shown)


samwillis (approved):
I think this is a good call. :shipit:

PR #387, opened 2026-06-13
feat(server-rust): high-performance Rust server implementation

balegas:
## What

A Rust implementation of the Durable Streams server (`packages/server-rust`) — a single self-contained binary with no database, broker, or other moving parts, just a process and a data directory. It stores each stream as the literal bytes it puts on the wire, so a catch-up read is a byte range of a file. The full core conformance suite passes, in **every supported run configuration**.

It ships **a single, hand-rolled HTTP/1.1 engine** — no framework. It owns the socket, so on Linux it serves reads with **`sendfile(2)`** (zero-copy page cache → socket) and binary appends with **`splice(2)`**; elsewhere it falls back to positioned reads. (`sendfile`/`splice` are always-permitted syscalls, so the zero-copy path works even under restrictive container seccomp.) Request *parsing* is delegated to the fuzzed `httparse` — only the framing decisions (keep-alive, chunked, Content-Length vs Transfer-Encoding) are bespoke, with the classic CL/TE request-smuggling vectors explicitly rejected; a panic in one connection drops only that connection.

> This started as three interchangeable engines (`hyper`, `raw`, `io_uring`) behind a `--http-engine` flag and was **collapsed to the one `raw` engine** (PR #388): `io_uring` won small reads on CPU-constrained hosts but lost on large reads, appends, and CPU efficiency — one zero-copy engine is the better product.

## Durability — two modes (`--durability`)

Both modes coexist in one binary and share the read path, store, and tiering — only the append path differs.

- **`wal` (default) — single-node crash durability.** Every append is written to the per-stream data file (the read surface) and staged into one of N sharded write-ahead logs; a per-shard group-commit committer `fdatasync`s the segment and only then acks. One committer batches **many streams' appends into a single fat fsync**, so the server is cardinality-insensitive (as fast on 10,000 streams as on 10). On boot, recovery replays the WAL, repairs any torn tail, and every record carries a header CRC + a **payload CRC32C verified on recovery**, so no torn record is ever served.
- **`memory` (Linux-only) — maximum throughput, durability via replication.** No WAL, no `fsync`: binary appends move `socket→file` via `splice(2)`, JSON appends are buffered writes, and the ack fires on the page-cache write. Recovery is the sidecar pass (rebuild from the per-stream files + `.meta`). **Not locally crash-durable, by design** — a power loss can lose un-written-back pages; durability is delegated to a future replication layer. Rejects on non-Linux.

Visibility and durability are decoupled in both modes: bytes land in the page cache and the tail is published (waking live readers) before durability resolves, so subscribers see an append at memory latency while the appender's 2xx waits for the durability guarantee.

## Main optimizations

1. **Contiguous wire-byte storage.** The data file holds exactly the bytes the protocol puts on the wire — JSON as `value,` runs, binary as raw payload. Every catch-up read is a literal byte range of the file. No length-prefix framing, no per-message reassembly. This is what makes `sendfile`/`splice` possible and restart recovery trivial (tail = base offset + file size).
2. **Group-commit coalesced fsync** (`wal` mode). Every append acks only after a covering fsync, but concurrent appenders share one in-flight barrier, so N concurrent appends cost ~1 fsync.
3. **Per-stream single-writer serialization, no global locks.**
4. **Lock-free read path** with a resident tail-chunk cache shared across subscribers.
5. **Event-driven live modes** (long-poll / SSE park on a per-stream watch channel).
6. **Zero-copy I/O** — `sendfile` for reads, `splice` for binary appends (straight to the per-stream file in `memory` mode).

## Cold-storage tiering (S3-compatible, opt-in)

`--tier s3|local` offloads sealed, immutable segments to any S3-compatible endpoint (R2, Tigris, MinIO, B2, …), keeping the hot tail local. Lifecycle: seal → upload → `head`-verify → durably flip `local → remote` → unlink the staged chunk file. Reads resolve each offset against a per-stream manifest: local ranges keep the zero-copy path; remote ranges stream as a bounded `Body::Channel` (one range-GET per segment). The live file's redundant sealed prefix is reclaimed by crash-safe **compaction** (rewrite to the hot tail under the appender lock; in-flight reads drain off the old fd). Feature-gated (`--features tier`); a default build keeps zero extra deps.

**Cold-read error surfacing.** A backend error or a short/truncated object on a streamed (cold/remote or large-local) read **aborts the connection** rather than completing a clean-but-truncated chunked `200` — so a client never resumes past `stream-next-offset` and silently skips a gap. (`api::StreamBody` carries a `failed` flag; the engine aborts on it. Found by a deterministic-simulation harness; regression-tested.)

## Run-configuration conformance matrix

CI runs the whole conformance suite **once per supported run configuration**, so every way of running the server stays protocol-equivalent: `wal-default`, `wal-resident-cache` (`--tail-cache-bytes`), `wal-read-offload-always`, and `memory` (`--durability memory`). Each passes the full core suite identically.

## Observability (opt-in, zero-cost when off)

`--features telemetry` adds OpenTelemetry (OTLP) spans/metrics on the bottleneck paths — zero overhead when off (ZST timers, `cfg`-gated counters).

## Scope

Full core protocol: create/append/read (catch-up, long-poll, SSE), HEAD, DELETE, JSON mode, idempotent producers, Stream-Seq, close, TTL/expiry, cursors, ETags/304, security headers, **stream forks**, and **restart recovery** via `.meta` sidecars.

**Not implemented: subscriptions / the `__ds` control plane.** Webhook and pull-wake subscriptions (Ed25519-signed delivery, JWKS) are out of scope for this server — clients consume streams directly via long-poll / SSE; run push delivery as a separate component on top. The core protocol conformance suite passes; the subscription tests are out of scope.

**Distribution.** Released through three channels off one git tag — `cargo` (the `durable-streams` crate), `npm` (`@durable-streams/server-rust` + per-platform prebuilt binaries), and GitHub Releases tarballs — via OIDC trusted publishing.

## Benchmarks

Dedicated 12-core Xeon (Linux 6.8): the server runs in its own cgroup and `wrk` is `taskset`-pinned to disjoint cores (a reserved core keeps `sshd` schedulable), governor `performance`, 3 repeats per cell.

**Reads** (conn 256): 1 KB **236k /s** (508% CPU), 16 KB 160k /s (456%), 1 MB 11.2k /s (266%). Read scaling (1 KB): 2c **193k**, 4c **256k**, 8c 236k /s — scales until the 3-core load generator saturates past 4 server cores. Hot reads stay sub-millisecond (p50 ≤ 0.11 ms) even under a 512 MB-capped cold backfill.

**Appends** (100 B, group commit): 116k /s @ conn 64, **210k /s** @ conn 256. Cold-tier read (`--tier local`): ~**5 GB/s**.

**vs [Ursula](https://github.com/tonbo-io/ursula)** (Raft-based durable-streams server; both pure local-durable, pinned to the same 8 cores): reads **1.8–3×** in our favour at comparable-or-less CPU; at an equal CPU budget (2c/4c) **3–4×**. Concurrent single-record appends are dominated by group commit (219k vs 478 /s @ conn 256 — Ursula ~one fsync/req). Fair bulk ingest (Ursula `append-batch` vs a single equal-size POST, one fsync each): **287 vs 180 MB/s (1.6×) at a third of the CPU**.

**Byte vs JSON mode** (100 B): JSON single-value append ≈ binary (fsync / group-commit bound), JSON reads ~6% below binary; 10-record array POSTs sustain ~2.1M records/s.


balegas:
**Progress update — pluggable HTTP engine**

- `1bc73c6e` refactor: handlers now speak engine-agnostic `Req`/`Resp`/`Body` types. Read responses are described as `Body::FileRange` (data-file segments + JSON framing bytes); each engine decides how to serve them. hyper becomes an adapter (`engine_hyper.rs`). No behavior change — conformance 326/0.
- `89a1a18a` feat: custom HTTP/1.1 engine (`engine_raw.rs`, tokio + httparse) selectable via `--http-engine raw` (default remains `hyper`). Keep-alive, content-length + chunked request bodies, `Expect: 100-continue`, chunked streaming for SSE. **Both engines pass the full conformance suite (326/0).**

Next: sendfile(2) zero-copy reads in the raw engine on Linux, then the `__ds` control plane (subscriptions) for full-suite coverage, then final benchmarks across engines.

balegas:
**Progress update — zero-copy reads + full protocol coverage**

- `bb904626` feat: **sendfile(2) zero-copy reads** in the raw engine on Linux — `Body::FileRange` segments are sent page-cache → socket by the kernel with no userspace buffer (write-readiness driven on the nonblocking socket). Validated in a Linux container: conformance 326/0 with `--http-engine raw`, including fork-chain responses stitched from multiple sendfile segments with JSON framing interleaved.
- `1bef5f77` feat: **`__ds` control plane** — webhook + pull-wake subscriptions with glob patterns, explicit stream membership, Ed25519-signed deliveries + JWKS, synchronous done auto-ack, async callbacks, claim/ack/release leases with generation fencing, SSRF URL validation.

**The full conformance suite now passes with nothing skipped: 332 / 332 — on both HTTP engines (hyper and raw).**

Remaining: final benchmark sweep (Node vs rust-hyper vs rust-raw, macOS + Linux) and the results tables in the PR body.

balegas:
**Progress update — review passes, bug fixes, cleanup**

Two code-review passes (correctness / cleanup / performance), fixes applied and re-verified:

**Bugs fixed**
- `engine_raw` sendfile: a 0-byte `sendfile` return with bytes still pending now errors instead of spinning forever (e.g. file truncated mid-read); zero-length segments short-circuit.
- `engine_raw` chunked request decode + `subs::decode_chunked` (webhook responses): `checked_add` bounds guarding against hostile/truncated chunk-size lines. The webhook one was a real panic (`&raw[size+2..]` without trailing CRLF) — with `panic=abort` that's a webhook-response-triggered crash.

**Cleanup (no behavior change)**
- Shared base64 (charset + pad) replaces two near-identical encoders.
- `store::materialize_segments()` consolidates the buffered segment-read duplicated across the hyper engine, SSE, and inline reads.
- `Segment::file_end()` and `subs::stream_list_json()` remove repeated arithmetic / payload-building. clippy now clean.

**Performance**
- The two constant security headers are emitted by each engine's writer instead of allocating two Strings per response. Post-change scale-out (raw, macOS): **33,020 appends/s, 3,999 MB/s reads** — no regression.

The second pass also confirmed the control plane holds **no std `Mutex` across any `.await`**, and the fsync-coalescer / generation-fencing / lease-expiry logic is race-free.

Both HTTP engines remain **332 / 332** on the full conformance suite.

AntonOfTheWoods:
@balegas , the latest releases still came from https://github.com/electric-sql/electric - is this PR going to be merged in some form?

arrowcircle:
Any updates on when this will be merged?

PR #392, opened 2026-07-10
Add @durable-streams/server-cloudflare: protocol server on Workers + Durable Objects, with a chat-cloudflare example

tombeckenham:
## What

**`packages/server-cloudflare`** — a conformant Durable Streams protocol server on Cloudflare Workers + Durable Objects, consumable as a library you mount in your own Worker. Imported from [durable-object-streams](https://github.com/tombeckenham/durable-object-streams) with full git history (rewritten under `packages/server-cloudflare/` via git-filter-repo).

- Every stream is its own DO instance (`idFromName(streamPath)`): per-stream serialization and durability-before-ack come from the DO input/output gates; state lives in DO SQLite; TTLs use DO alarms; fork semantics (refcounted soft-delete, stitched reads, cascade GC) work across objects via DO-to-DO RPC.
- Pure library package: no wrangler.jsonc, no dev server, no bundled example. Runtime types come from `wrangler types` (generated `worker-configuration.d.ts` is checked in). Lint/format come from the repo root.
- Conformance is wired the same way as caddy-plugin: `test/conformance.test.ts` boots the consumer template via `wrangler dev` (test-only config) and runs the shared `@durable-streams/server-conformance-tests` suite, registered as the `server-cloudflare` vitest project. **326 passed, 0 failed, 6 skipped** — identical to the reference server (skips are the suite's subscriptions-gated webhook tests; the experimental `__ds` control plane is out of scope).

**`examples/chat-cloudflare`** — the chat-tanstack app rebuilt as a single Cloudflare Worker with the Cloudflare Vite plugin (TanStack Start + the streams server + the DO in one deployable), including a Deploy to Cloudflare button.

- Server routes call the streams handler **in-process** against the `STREAMS` DO binding — no HTTP hop, and no "Worker cannot fetch its own hostname" issue in production. Deploy is `pnpm run deploy` + an `OPENAI_API_KEY` secret.
- Generation is kept alive past the `202` with `waitUntil` from `cloudflare:workers` (workerd cancels orphaned promises when the request context ends — the Node example tolerates this, Workers don't).
- Sidebar chat metadata is itself a durable stream (`chats/index`, last-write-wins by id), since Workers have no filesystem.
- One process dev: `pnpm dev` runs vite with workerd and a local DO.

**`@durable-streams/tanstack-ai-transport`** — adds a `fetchClient` option to `DurableStreamTarget` and `materializeSnapshotFromDurableStream`, passed through to the client's existing custom-fetch support. This is what lets server-side helpers run against an in-process handler or a service binding. Patch changeset included.

## Repo plumbing

- Root tsconfig excludes `packages/server-cloudflare` (Workers types conflict with the DOM sweep, same treatment as caddy-plugin); root `typecheck` runs the package's own typecheck.
- Root eslint anchors `tsconfigRootDir` so typed linting works regardless of the ESLint server's working directory, and ignores build artifacts (`.next`, `.nitro`, `.tanstack`, `.wrangler`, `worker-configuration.d.ts`).
- `.tanstack/` gitignored globally (was a yjs-demo-specific entry).
- New root `dev:libs` script: watch `packages/*` without starting every example/docs server.
- server package quiets SSE pump warnings for routine client disconnects (workerd's "Network connection lost.").

## Verification

- Full server conformance suite against `wrangler dev` (workerd, local DO SQLite): 326 passed / 6 skipped, run via `vitest run --project server-cloudflare`.
- Example verified end-to-end in dev with a real OpenAI key: chat creation, generation through `RUN_FINISHED`, stream-backed sidebar index, SSE read proxy, resume after restart (DO SQLite persistence).
- Root typecheck, build, and lint green on the new package and example.


KyleAMathews:
### Code review

Thanks for the substantial implementation and conformance coverage. I found 4 issues that should be addressed before merging; each was reproduced with a failing test:

1. **Close-only POSTs do not refresh sliding TTL.** Both the plain and producer-aware close paths let a stream expire at its original deadline instead of renewing its idle TTL. A focused workerd test failed for both paths: after closing, both streams returned `404` at the original expiry. Adding `touchAccess(now)` and `await syncExpiryAlarm()` to both successful branches made both cases pass.

https://github.com/durable-streams/durable-streams/blob/f6d02b46eeb542ff1b62c101edac17d007d2fcce/packages/server-cloudflare/src/stream-object.ts#L1374-L1444

2. **CORS preflight does not allow `If-None-Match`.** The new conditional-request preflight test failed against the in-memory and file-backed servers because `Access-Control-Allow-Headers` omitted it. After adding the header to the relevant lists, the targeted server and Cloudflare runs passed (`3 passed`). Please include shared conformance coverage with the fix.

https://github.com/durable-streams/durable-streams/blob/f6d02b46eeb542ff1b62c101edac17d007d2fcce/packages/server-cloudflare/src/stream-object.ts#L1862-L1875

3. **Authenticated catch-up responses are publicly cacheable.** A behavioral test using configured custom auth failed because the authenticated response returned `Cache-Control: public, max-age=60, stale-while-revalidate=300`. Rewriting authenticated responses to `no-store` made the focused test pass. A configurable private/public policy is also reasonable, but authenticated responses should not default to shared public caching.

4. **`cors: false` does not apply to actual stream responses.** A handler test failed because a forwarded response still contained `Access-Control-Allow-Origin: *`: the handler disables its own CORS headers, but `StreamObject.respond()` adds them unconditionally. Stripping those headers when CORS is disabled made the test pass.

https://github.com/durable-streams/durable-streams/blob/f6d02b46eeb542ff1b62c101edac17d007d2fcce/packages/server-cloudflare/src/handler.ts#L78-L124

https://github.com/durable-streams/durable-streams/blob/f6d02b46eeb542ff1b62c101edac17d007d2fcce/packages/server-cloudflare/src/stream-object.ts#L1856-L1877

There is also a security/design concern in the deployable chat example: it publicly mounts unauthenticated `/streams/*` routes, and an unauthenticated `PUT /streams/probe` reached the protocol handler and returned `201` in a focused test. Chat IDs are also truncated to 8 hex characters (32 bits). If this example is intended to be safely deployable, raw protocol access should be protected or kept internal and full UUIDs should be used; otherwise the insecure demo posture should be documented prominently.

https://github.com/durable-streams/durable-streams/blob/f6d02b46eeb542ff1b62c101edac17d007d2fcce/examples/chat-cloudflare/src/server.ts#L12-L19

https://github.com/durable-streams/durable-streams/blob/f6d02b46eeb542ff1b62c101edac17d007d2fcce/examples/chat-cloudflare/src/lib/chat-store.ts#L55-L63

The `subscribe()`/`fetchClient` inconsistency is real, but merge-base verification showed it predates this PR, so it should be tracked separately rather than blocking this change.

tombeckenham:
All five points addressed, thanks for the thorough review:

1. **Close-only TTL refresh** — fixed in `handleCloseOnly` (both the plain and producer paths now `touchAccess` + re-arm the expiry alarm). The same gap existed in the reference TS server and the Caddy plugin's memory/file stores, so those are fixed too, with two new shared conformance tests (`should extend TTL on close-only POST` / `producer close-only POST`). (92958132)

2. **Preflight `If-None-Match`** — added to `Access-Control-Allow-Headers` in the Worker handler, the DO responses, and the reference server (Caddy already allowed it), with a shared conformance test for the preflight. (92958132)

3. **Authenticated responses no longer publicly cacheable** — when an auth gate is active (custom `auth` hook or `AUTH_TOKEN` set), the handler rewrites publicly-cacheable catch-up responses to `Cache-Control: no-store`. (6a6c6248)

4. **`cors: false` now applies to stream responses** — the handler strips the DO-stamped `Access-Control-*` headers from forwarded responses when CORS is disabled. Both 3 and 4 are covered by new handler unit tests (they're package option behavior rather than protocol behavior, so they didn't fit the shared suite). (6a6c6248)

5. **Chat example** — the public unauthenticated `/streams/*` mount is gone (the app never needed it: server routes call the handler in-process and the browser reads via the `/api/chat-stream` proxy), chat ids are full UUIDs, and the README documents how to expose the protocol behind auth if wanted. (0313a92f)

Agreed on tracking the pre-existing `subscribe()`/`fetchClient` inconsistency separately.

tombeckenham:
@KyleAMathews all five review points are addressed and pushed (summary above) — ready for another look when you have a moment.

KyleAMathews:
## Review findings: confirmed with failing regression tests

I evaluated the reported issues against this PR and reproduced all five as real bugs. The severity labels below reflect the correctness and durability impact observed in the tests.

### 1. High: JSON SSE catch-up batches produce invalid logical JSON

`pumpSse()` emits each buffered JSON message as a separate, complete JSON array in its own `event: data` frame, but emits only one trailing `event: control` frame for the whole read batch.

For two buffered messages, clients that collect data until the control boundary observe the equivalent of:

```text
[message1][message2]
```

That is not valid JSON. The data/control framing is also inconsistent: multiple data events share a single control boundary.

**Failing regression:** append at least two JSON messages before opening the SSE subscription, collect the catch-up data through its control event, and parse it as one JSON value. The pre-fix implementation fails because it yields concatenated arrays.

**Recommended fix:** flatten all JSON messages returned in a read batch with `formatJsonMessages()` and emit one JSON `data` event followed by one `control` event.

---

### 2. Critical: fork reference acquisition and release are not idempotent

Fork lifetime management currently crosses Durable Object transaction boundaries using a counter and a GC retry record that cannot uniquely identify a fork edge or generation. An RPC can commit remotely while its response is lost, causing the caller to retry an operation that already succeeded.

This creates two correctness failures:

- A repeated acquire can count the same fork twice and leak a soft-deleted source.
- A repeated release can decrement twice and purge a source while another child still depends on inherited data.

A path-only pending release also permits an old release to affect a newly recreated stream at the same path.

**Failing regressions:** three dedicated tests fail with the counter-based implementation:

1. **Repeated acquire:** the same acquire is applied twice. After release, the source incorrectly remains `410` rather than being purged to `404`.
2. **Release response loss:** the source commits a release, the response is lost, and the retained local intent retries. The retry consumes another live fork's reference, producing `404` instead of the expected `410`.
3. **Same-path generation reuse:** a delayed release belonging to an earlier generation decrements the recreated generation, producing `404` instead of `410`.

A further failure-injection test covers the acquire boundary: the source commits `forkAcquire`, its response is lost, and the child retries using its durable intent.

**Recommended fix:** assign every fork relationship a stable edge UUID; persist the child's edge intent before remote acquisition; persist source-side active edges keyed by edge ID; qualify edges and releases with the source generation; and make acquire an insert-if-absent and release a delete-if-present. Durable child intents should distinguish acquisition, acquired, and releasing states so retries reuse the same identity.

---

### 3. High: a consumed sliding-TTL alarm can fail to rearm

`syncExpiryAlarm()` suppresses `setAlarm()` when the desired timestamp is within 500 ms of the in-memory `lastArmedAlarm`. That cache can remain equal to an alarm that has already fired and been consumed.

A deterministic sequence is:

1. Arm expiry for `T`.
2. A read shortly before `T` moves expiry to `T + 100 ms`; rearming is suppressed by the 500 ms threshold.
3. The alarm fires at `T`, but the stream is not expired.
4. The handler asks to schedule `T + 100 ms`, compares it with stale `lastArmedAlarm === T`, and suppresses the new alarm.
5. No alarm remains to purge the stream.

The same stale-cache problem exists after explicit `deleteAlarm()` and can affect immediate delete/recreate behavior.

**Failing regression:** with alarm-entry cache clearing removed, the test advances a sliding expiry by less than 500 ms before the original alarm is consumed. After the alarm handler runs, storage reports no armed alarm (`null`) even though a future expiry duty remains.

**Recommended fix:** clear `lastArmedAlarm` when alarm execution begins and after every `deleteAlarm()`. Alarm synchronization should compare with actual storage alarm state and arm the earliest durable outstanding duty across stream expiry and GC retry.

---

### 4. Medium: non-live `GET ?offset=now` does not refresh sliding TTL

The non-live `offset=now` fast path returns before `touchAccess()` and `syncExpiryAlarm()`. The protocol's sliding TTL resets on reads and writes other than `HEAD`, so this valid GET currently does not count as access.

**Failing regression:** create a stream with a short sliding TTL, wait, issue only non-live `GET ?offset=now`, and then check after the original expiry but before the refreshed expiry. The stream is incorrectly purged under the pre-fix behavior.

**Recommended fix:** after request validation, touch access and synchronize the expiry alarm before taking the early response path.

---

### 5. High: the 4 MiB response budget resets at every fork segment

`readStitched()` can read a full budget from an inherited source and then give the child store a new full budget. Deeper fork chains can add another allowance per segment, so the nominal per-response and structured-clone bound scales with fork depth.

**Failing regression:** place data on both sides of a fork boundary with a combined size over the limit. The pre-fix one-level fork response contains **4,500,000 bytes**, exceeding the intended 4 MiB response budget.

**Recommended fix:** return or calculate bytes used by inherited reads and pass only the remaining budget to the local read and through parent RPCs. The oversized-first-message exception should apply only when the response has not already emitted inherited data. Continuation reads should recover the remainder without duplication.

---

## Verification summary

All five findings were demonstrated by failing tests against the affected behavior. With the corresponding fixes applied locally:

```text
339 tests passed
6 skipped
2 test files passed
```

The Cloudflare package typecheck, targeted formatting checks, and `git diff --check` also passed.

Given the critical fork-reference failure and the high-severity SSE, alarm, and response-budget defects, I recommend addressing these before merge.

tombeckenham:
All five findings confirmed and fixed in fe5c15ef — each is now covered by a regression test that fails without its fix:

1. **JSON SSE catch-up framing** — `pumpSse` now flattens the whole read batch with `formatJsonMessages()` into ONE `data` event followed by ONE `control` event (binary/text batches are likewise concatenated into a single event). The reference TS server had the same per-message framing and is fixed too; Caddy already batched correctly. Covered by a shared conformance test (`JSON SSE catch-up pairs every data event with a control event`) that asserts data/control alternation and that each data payload parses as a complete JSON value.

2. **Fork reference idempotency** — reimplemented as you recommended: references are now fork-edge rows with stable UUIDs (insert-if-absent acquire, delete-if-present release), releases are qualified by the source generation, and the child persists its edge intent durably *before* the remote acquire so a create retry reuses the same identity. Five direct-DO regression tests cover repeated acquire, retried release with another live fork, delayed release against a recreated generation, the lost-response-mid-create injection, and the idempotent PUT retry.

3. **Alarm rearm after consumption** — the in-memory `lastArmedAlarm` cache is cleared when the alarm fires and after `deleteAlarm`, and arming now never delays an earlier still-pending alarm (GC retries and expiry share one duty queue). Two direct-DO tests assert `storage.getAlarm()` is rearmed after a consumed alarm and after delete+recreate.

4. **`GET ?offset=now` TTL refresh** — the early-return path now touches access time in all three servers (Cloudflare, reference TS, Caddy — Caddy reuses a tail `store.Read`, which is how a normal caught-up GET refreshes). Covered by a shared conformance test (`should extend TTL on GET ?offset=now`).

5. **Fork read budget** — `readStitched` threads one byte budget through the whole chain: inherited segments spend from it via the `readRange` RPC, and the oversized-first-message exception only applies when the response is still empty. The regression test reproduces your 4,500,000-byte response and now asserts ≤ 4 MiB per response with byte-exact continuation (no loss or duplication across the capped boundary).

The direct-DO tests run on a new `vitest-pool-workers` harness (`packages/server-cloudflare/test-do/`, `pnpm test:do`) since alarms, RPC retries, and storage state aren't observable over HTTP. Also wired a CI job that runs both the server-cloudflare conformance suite and the DO tests — neither ran in CI before — and made the wrangler-dev conformance boot hermetic (`--persist-to` a fresh dir).

KyleAMathews (approved):
Thanks so much! And given the latest models are finding somewhat obscure issues now with older code haha we're probably getting into diminishing returns and should merge this.

PR #393, opened 2026-07-13
feat(yjs): add snapshot availability webhooks

icehaunter:
## Motivation

Yjs snapshots are currently an internal implementation detail. External persistence workers have no supported way to learn when the current snapshot advances without polling or coupling directly to internal stream layout.

## What changed

- Add service-scoped PUT, GET, and DELETE endpoints for `snapshot.available` webhook subscriptions.
- Translate document globs to Durable Streams subscriptions over each matching document's `.index` stream.
- Reuse the existing Durable Streams signing, retry, lease, callback, and acknowledgement machinery.
- Return the public subscription ID alongside the opaque delivery subscription ID used in signed webhook payloads.
- Add lifecycle, delivery, and invalid-event conformance coverage, and include the Yjs project in the repository test command.
- Document the API and add a patch changeset for `@durable-streams/y-durable-streams`.

## Design decisions

Webhook deliveries are wake notifications, not snapshot payloads. Receivers derive the document from each pending index stream, fetch `?offset=snapshot`, and persist the current binary snapshot before acknowledging completion.

The guarantee is latest-snapshot convergence: multiple compactions may coalesce, and acknowledgements retain pending work when a newer snapshot arrives during processing. New subscriptions observe future snapshots only; consumers that need the snapshot current at subscription creation bootstrap it separately.

This facade depends on an underlying Durable Streams implementation that supports the current reserved subscription API and webhook delivery.

### DIFF
--- a/packages/y-durable-streams/src/server/snapshot-subscriptions.ts
+++ b/packages/y-durable-streams/src/server/snapshot-subscriptions.ts
@@ -1,0 +1,320 @@
+import type { IncomingMessage, ServerResponse } from "node:http"
+
+const SNAPSHOT_AVAILABLE_EVENT = `snapshot.available`
+
+export interface SnapshotSubscriptionRoute {
+  service: string
+  subscriptionId: string
+}
+
+interface SnapshotSubscriptionInput {
+  type: `webhook`
+  events: Array<typeof SNAPSHOT_AVAILABLE_EVENT>
+  documentPattern: string
+  webhook: { url: string }
+  leaseTtlMs?: number
+  description?: string
+}
+
+export function parseSnapshotSubscriptionRoute(
+  path: string
+): SnapshotSubscriptionRoute | null {
+  const match = path.match(/^\/v1\/yjs\/([^/]+)\/__ds\/subscriptions\/([^/]+)$/)
+  if (!match) return null
+
+  try {
+    const service = decodeURIComponent(match[1]!)
+    const subscriptionId = decodeURIComponent(match[2]!)
+    if (!/^[a-zA-Z0-9_-]+$/.test(service)) return null
+    if (
+      subscriptionId.length === 0 ||
+      subscriptionId.length > 256 ||
+      subscriptionId.includes(`/`)
+    ) {
+      return null
+    }
+    return { service, subscriptionId }
+  } catch {
+    return null
+  }
+}
+
+export class SnapshotSubscriptionHandler {
+  private readonly dsServerUrl: string
+  private readonly dsServerHeaders: Record<string, string>
+
+  constructor(options: {
+    dsServerUrl: string
+    dsServerHeaders: Record<string, string>
+  }) {
+    this.dsServerUrl = options.dsServerUrl
+    this.dsServerHeaders = options.dsServerHeaders
+  }
+
+  async handle(
+    req: IncomingMessage,
+    res: ServerResponse,
+    route: SnapshotSubscriptionRoute
+  ): Promise<void> {
+    const method = req.method?.toUpperCase()
+    if (method === `PUT`) {
+      await this.createOrConfirm(req, res, route)
+      return
+    }
+    if (method === `GET`) {
+      await this.read(req, res, route)
+      return
+    }
+    if (method === `DELETE`) {
+      await this.delete(req, res, route)
+      return
+    }
+
+    this.writeError(res, 405, `INVALID_REQUEST`, `Method not allowed`)
+  }
+
+  private async createOrConfirm(
+    req: IncomingMessage,
+    res: ServerResponse,
+    route: SnapshotSubscriptionRoute
+  ): Promise<void> {
+    let body: unknown
+    try {
+      body = JSON.parse((await this.readBody(req)).toString(`utf8`))
+    } catch {
+      this.writeError(res, 400, `INVALID_REQUEST`, `Invalid JSON body`)
+      return
+    }
+
+    const parsed = this.parseInput(body)
+    if (`error` in parsed) {
+      this.writeError(res, 400, `INVALID_REQUEST`, parsed.error)
+      return
+    }
+
+    const input = parsed.value
+    const response = await fetch(this.subscriptionUrl(route), {
+      method: `PUT`,
+      headers: this.requestHeaders(req, true),
+      body: JSON.stringify({
+        type: input.type,
+        pattern: this.indexPattern(route.service, input.documentPattern),
+        webhook: input.webhook,
+        lease_ttl_ms: input.leaseTtlMs,
+        description: input.description,
+      }),
+    })
+
+    await this.forwardResponse(res, response, route)
+  }
+
+  private async read(
+    req: IncomingMessage,
+    res: ServerResponse,
+    route: SnapshotSubscriptionRoute
+  ): Promise<void> {
+    const response = await fetch(this.subscriptionUrl(route), {
+      method: `GET`,
+      headers: this.requestHeaders(req),
+    })
+    await this.forwardResponse(res, response, route)
+  }
+
+  private async delete(
+    req: IncomingMessage,
+    res: ServerResponse,
+    route: SnapshotSubscriptionRoute
+  ): Promise<void> {
+    const response = await fetch(this.subscriptionUrl(route), {
+      method: `DELETE`,
+      headers: this.requestHeaders(req),
+    })
+    await this.forwardResponse(res, response, route)
+  }
+
+  private parseInput(
+    body: unknown
+  ): { value: SnapshotSubscriptionInput } | { error: string } {
+    if (!body || typeof body !== `object`) {
+      return { error: `Request body must be a JSON object` }
+    }
+
+    const input = body as Record<string, unknown>
+    if (input.type !== `webhook`) {
+      return { error: `type must be webhook` }
+    }
+    if (
+      !Array.isArray(input.events) ||
+      input.events.length !== 1 ||
+      input.events[0] !== SNAPSHOT_AVAILABLE_EVENT
+    ) {
+      return { error: `events must contain only snapshot.available` }
+    }
+
+    const documentPattern = this.normalizeDocumentPattern(
+      input.document_pattern
+    )
+    if (!documentPattern) {
+      return { error: `document_pattern must be a valid document glob` }
+    }
+
+    const webhook = input.webhook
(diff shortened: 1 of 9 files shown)


packages/y-durable-streams/src/server/snapshot-subscriptions.ts
```diff
@@ -0,0 +1,320 @@
...
+    for (const segment of segments) {
+      if (segment === `*` || segment === `**`) continue
+      if (!/^[a-zA-Z0-9_-]+$/.test(segment)) return null
+    }
+    return normalized
+  }
+
+  private indexPattern(service: string, documentPattern: string): string {
+    return `yjs/${service}/docs/${documentPattern}/.index`
+  }
```
balegas, line 222: **Correctness: this pattern also matches awareness index streams, producing spurious `snapshot.available` wakes.**

The DS glob's `*`/`**` match dot-prefixed segments, so the composed pattern matches more than snapshot index streams. Verified against `packages/server/src/glob.ts`:

```js
globMatch('yjs/svc/docs/projects/**/.index', 'yjs/svc/docs/projects/foo/.awareness/.index')  // → true
globMatch('yjs/svc/docs/projects/*/.index',  'yjs/svc/docs/projects/.awareness/.index')      // → true
```

Creating a non-default awareness stream appends to `.../docs/{doc}/.awareness/.index` (see `appendToIndexStream` in `yjs-server.ts`), and `SubscriptionManager.onStreamAppend` glob-links and wakes on any matching append. So a presence-enabled app under a subscribed pattern fires `snapshot.available` webhooks with no snapshot involved.

This is worse than noise: per the documented receiver algorithm (YJS-PROTOCOL §5.11), the receiver derives the doc path — here `projects/foo/.awareness` — and fetches `?offset=snapshot`, which 400s (`PathUtils.normalize` rejects dot segments). A receiver that follows "return `{done: true}` only after persistence succeeds" never acks that stream and retries forever.

Options:
- Since valid document path segments can never contain dots, filter/exclude these matches (e.g. rename the awareness index so it doesn't end in `/.index`, or extend the glob to express "segments without a leading dot").
- At minimum, document that receivers MUST ignore (and ack as done) pending index streams whose derived document path is not a valid document path, and add a conformance test covering an awareness stream created under a subscribed pattern.

packages/y-durable-streams/src/server/snapshot-subscriptions.ts
```diff
@@ -0,0 +1,320 @@
...
+    }
+
+    const parsed = this.parseInput(body)
+    if (`error` in parsed) {
+      this.writeError(res, 400, `INVALID_REQUEST`, parsed.error)
+      return
+    }
+
+    const input = parsed.value
+    const response = await fetch(this.subscriptionUrl(route), {
```
balegas: **Robustness: unhandled `fetch` failure surfaces as `500 INTERNAL_ERROR` instead of `502 PROXY_ERROR`.**

These upstream calls (here and in `read`/`delete`) have no try/catch, so if the DS server is unreachable the error bubbles to the top-level handler and becomes a generic 500. The existing `proxyToDsServer` maps the same condition to `502 PROXY_ERROR` — wrapping these for consistency would make DS-server outages distinguishable from facade bugs.

packages/y-durable-streams/src/server/snapshot-subscriptions.ts
```diff
@@ -0,0 +1,320 @@
...
+export function parseSnapshotSubscriptionRoute(
+  path: string
+): SnapshotSubscriptionRoute | null {
+  const match = path.match(/^\/v1\/yjs\/([^/]+)\/__ds\/subscriptions\/([^/]+)$/)
+  if (!match) return null
+
+  try {
+    const service = decodeURIComponent(match[1]!)
+    const subscriptionId = decodeURIComponent(match[2]!)
+    if (!/^[a-zA-Z0-9_-]+$/.test(service)) return null
```
balegas: **Consistency: service validation is stricter than the document routes.**

`parseRoute` accepts any `[^/]+` as the service name, but this requires `[a-zA-Z0-9_-]+`. A service like `my.svc` can host documents, yet its subscription URL falls through to the generic "Invalid document path" 400. Either align the two validations or note the restriction in YJS-PROTOCOL §5.11.

packages/y-durable-streams/src/server/snapshot-subscriptions.ts
```diff
@@ -0,0 +1,320 @@
...
+    const id = Buffer.from(route.subscriptionId).toString(`base64url`)
+    return `yjs:${service}:${id}`
+  }
+
+  private requestHeaders(
+    req: IncomingMessage,
+    jsonBody: boolean = false
+  ): Record<string, string> {
+    const headers: Record<string, string> = { ...this.dsServerHeaders }
+    for (const [key, value] of Object.entries(req.headers)) {
```
balegas, line 266: **Consider letting `dsServerHeaders` win on this endpoint.**

Client-supplied headers overwrite the configured `dsServerHeaders` here (e.g. a caller can override the `authorization` header used to reach the DS server). This mirrors `proxyToDsServer`, so it's consistent with existing behavior — but this endpoint targets the DS reserved subscription API, which is more privileged than a document proxy. Spreading `dsServerHeaders` last (or only forwarding an allowlist of client headers) would be safer.

packages/y-durable-streams/src/server/snapshot-subscriptions.ts
```diff
@@ -0,0 +1,320 @@
...
+    subscription: Record<string, unknown>,
+    service: string
+  ): string {
+    const pattern = subscription.pattern
+    if (typeof pattern !== `string`) return `**`
+    const prefix = `yjs/${service}/docs/`
+    const suffix = `/.index`
+    if (!pattern.startsWith(prefix) || !pattern.endsWith(suffix)) return `**`
+    return pattern.slice(prefix.length, -suffix.length)
+  }
```
balegas, line 362: **Nit: the `**` fallback silently misreports scope.**

If the upstream pattern doesn't fit the expected `yjs/{service}/docs/…/.index` shape, GET responds with `document_pattern: "**"`, which claims broader coverage than the subscription actually has. Omitting the field (or returning the raw upstream pattern) would be more honest for subscriptions created out of band.

packages/y-durable-streams/test/yjs-conformance.test.ts
```diff
@@ -237,6 +305,182 @@ describe(`Yjs Durable Streams Protocol`, () => {
...
+        expect(created.webhook.signing.jwks_url).toContain(
+          `/v1/stream/__ds/jwks.json`
+        )
+
+        const confirmedResponse = await fetch(subscriptionUrl, {
+          method: `PUT`,
+          headers: { "content-type": `application/json` },
+          body: JSON.stringify(request),
+        })
+        expect(confirmedResponse.status).toBe(200)
```
balegas, line 363: **Test coverage: the documented 409 path isn't exercised.**

YJS-PROTOCOL §5.11 says reusing the same ID with a different configuration returns `409 Conflict`, but the lifecycle test only re-confirms an identical config. Since conflict detection happens in the DS server *after* the facade rewrites the pattern, a test here would also pin the facade's normalization — e.g. re-PUT with `document_pattern: "/projects/**"` should still be 200 (normalizes to the same pattern), while a different `webhook.url` should be 409. A rejected-`document_pattern` case (invalid characters) would also be cheap to add.

balegas (commented):
## Review summary

Solid feature with the right architecture: the snapshot webhook endpoint is a thin facade over the existing Durable Streams subscription machinery (signing, retries, leases, acks all reused), the wake-notification semantics with latest-snapshot convergence are the right call, and the namespaced `delivery_subscription_id` cleanly avoids collisions with reserved DS subscriptions. Docs are thorough and honest about the semantics (coalescing, future-snapshots-only), and the tests are real end-to-end — actual HTTP webhook receiver, signature-format assertion, snapshot fetch through documented discovery. I ran the `y-durable-streams` test project locally: 64/64 pass, including the three new tests.

### Main issue (should fix before merge)

**Awareness index streams trigger spurious `snapshot.available` wakes.** The composed glob `yjs/{svc}/docs/{pattern}/.index` also matches `.../docs/{doc}/.awareness/.index`, because the DS glob's `*`/`**` match dot-prefixed segments (verified against `packages/server/src/glob.ts`). Creating a non-default awareness stream appends to that index, so presence activity fires snapshot webhooks. Worse, a receiver following the documented algorithm derives an invalid document path (`{doc}/.awareness`), gets a 400 from `?offset=snapshot`, never acks, and retries forever. Details and suggested fixes in the inline comment on `indexPattern`.

### Smaller items (inline)

- Unhandled upstream `fetch` failures return `500 INTERNAL_ERROR`; `proxyToDsServer` returns `502 PROXY_ERROR` for the same condition.
- Service-name validation on the subscription route (`[a-zA-Z0-9_-]+`) is stricter than document routes (`[^/]+`).
- Client headers override the configured `dsServerHeaders` (including auth) on this privileged endpoint.
- GET falls back to reporting `document_pattern: "**"` when the upstream pattern doesn't parse, overstating scope.
- The documented `409 Conflict` path isn't covered by the lifecycle test.

None of the smaller items are blocking; the awareness false-positive is the one that needs a resolution (code fix, or documented receiver-side filtering plus a conformance test) before this ships.

icehaunter:
Addressed all review feedback in `074e7566`:

- Prevented awareness activity from producing false `snapshot.available` wakes by moving new awareness bookkeeping to a sibling `.awareness-index` stream, retaining dual-read/delete support for the legacy `.awareness/.index` stream, and reserving invalid awareness names such as `.index`.
- Added `502 PROXY_ERROR` handling for Durable Streams connection failures.
- Made configured `dsServerHeaders` override caller headers case-insensitively and stripped hop-by-hop headers, including chunked request framing.
- Added support and documentation for dotted service names while rejecting percent-encoded spellings that would not match document routing.
- Removed the fabricated `document_pattern: "**"` fallback when an upstream scope cannot be decoded safely.
- Expanded conformance coverage for normalized re-confirmation, `409 Conflict`, invalid patterns, awareness non-delivery, configured auth, chunked proxying, outages, dotted services, malformed upstream scopes, and legacy awareness cleanup.

The final Yjs suite passes 68/68, the repository test suite passes 1,158 tests with 3 skipped, and all PR checks are green. Thanks for the careful review.

balegas (approved):

